PR #3, opened 2016-09-12
Link against conda-forge's libuuid

ceholden:
Goal is to fix linking issues with Qt apps when libraries are installed that use conda-forge's libuuid (e.g., gdal) are also installed.

See issue:
https://github.com/conda-forge/libuuid-feedstock/issues/8

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -30,7 +30,7 @@ source:
     - vs2015_int_fix.patch  # [win and py >= 35]
 
 build:
-  number: 0
+  number: 1
   skip: true  # [win or osx]
   features:
     - vc9  # [win and py27]
@@ -52,6 +52,8 @@ requirements:
     - jom  # [win]
     # This is here because conda is not pulling libxml2 from conda-forge
     - libxml2 >=2.9.3
+    # This is here because some conda-forge libraries link against conda-forge's libuuid
+    - libuuid
   run:
     - freetype 2.6.*  # [unix]
     - fontconfig 2.11.*  # [linux]
@@ -63,6 +65,8 @@ requirements:
     - zlib 1.2.*
     # This is here because conda is not pulling libxml2 from conda-forge
     - libxml2 >=2.9.3
+    # This is here because some conda-forge libraries link against conda-forge's libuuid
+    - libuuid
 
 test:
   files:


ccordoba12:
Thanks a lot @ceholden, but you forgot to bump the build number. I'm going to stop the build, so you can do that in a new commit :-)

ceholden:
Arg! I did forget. Thanks for catching that! I've updated to `number: 1`

Back to the beginning of the build again...

ceholden:
It passed :tada:! Just took 1 hour 21 minutes...

ccordoba12:
Ok, merging then :-) Please let me know if it fixes the problem for you.

ceholden:
@ccordoba12 These low level dependencies are like a virus infecting everything!

I'm not seeing any conflict with Qt and GDAL anymore because they're both using the same `libuuid`, but when I tried to build `qscintilla2` (using [this recipe](https://github.com/conda-forge/staged-recipes/pull/1395)), I get the following on import of `PyQt4.Qsci`:

`ImportError: .../conda-bld/._1473729027808/_t_env/lib/python2.7/site-packages/PyQt4/QtGui.so: undefined symbol: _ZN15QSessionManager16staticMetaObjectE`

The `qscintilla2` recipe pulls in PyQt4 from defaults which I think is now incompatible with the new build of conda-forge's Qt. To test, I compiled PyQt4 using your recipe in `conda/conda-recipes` against this new Qt build that uses conda-forge's `libuuid`. Using Qt4 from conda-forge and this local build of PyQt4, I was then able to successfully build `qscintilla2` and it passed its tests. I was also able to compile QGIS, the source of my reason for integrating Qt and GDAL, under these circumstances and it works beautifully.

This situation is also affecting the `qtconsole` feedstock, and presumably other Qt4 + PyQt4 feedstocks:
``` bash
> conda create -n test -c conda-forge qtconsole
> source activate test
> ipython qtconsole
...
ImportError: /data/conda/conda2/envs/test/lib/python2.7/site-packages/PyQt4/QtGui.so: undefined symbol: _ZN15QSessionManager16staticMetaObjectE
```

When I recreated the environment with the PyQt4 I compiled against conda-forge's Qt (created in this PR), everything was fine. 
``` bash
> conda create -n test -c conda-forge -c ceholden qtconsole
> source activate test
> ipython qtconsole  # success
```

I don't know how far the rabbit hole goes, but I'd be happy to work up a PR to add your `conda/conda-recipes` PyQt4 recipe to conda-forge. For the short term, it looks like we should pull this build somehow or recompile without `libuuid` so that PyQt4 from defaults can be compatible.

ccordoba12:
Ok, this means we just need to create a `pyqt` feedstock, right? Then everything Qt-related will be in the realm of conda-forge and we will avoid these ugly conflicts with defaults ;-)

ccordoba12:
And before you jump to create a PR for pyqt, let me do that using our latest Anaconda recipes.

ceholden:
Yes, I think that's the ideal. Happy to defer to you on the PyQt4 feedstock PR :).

Thanks very much for your attention to this!

PR #9, opened 2016-10-12
Add session management support for Qt

### LINKED ISSUE #4: QSessionManager linux-64/qt-4.8.7-1
building some conda packages with qt dependencies gives me this error:
```
invalid use of incomplete type ‘class QSessionManager’ manager.cancel();
```
 was there a change in the qt library? or is this related to the latest conda-forge qt build?

ccordoba12:
Fixes #4

### DIFF
--- a/ci_support/run_docker_build.sh
+++ b/ci_support/run_docker_build.sh
@@ -46,7 +46,7 @@ source run_conda_forge_build_setup
 # "recipe/yum_requirements.txt" file. After updating that file,
 # run "conda smithy rerender" and this line be updated
 # automatically.
-yum install -y libX11-devel libXt-devel libXext-devel chrpath libXrender-devel gtk2-devel dbus-devel
+yum install -y libX11-devel libXt-devel libXext-devel chrpath libXrender-devel gtk2-devel dbus-devel libSM libICE
 
 
 # Embarking on 1 case(s).
--- a/ci_support/run_docker_build.sh
+++ b/ci_support/run_docker_build.sh
@@ -46,7 +46,7 @@ source run_conda_forge_build_setup
 # "recipe/yum_requirements.txt" file. After updating that file,
 # run "conda smithy rerender" and this line be updated
 # automatically.
-yum install -y libX11-devel libXt-devel libXext-devel chrpath libXrender-devel gtk2-devel dbus-devel libSM libICE
+yum install -y libX11-devel libXt-devel libXext-devel chrpath libXrender-devel gtk2-devel dbus-devel libSM-devel libICE-devel
 
 
 # Embarking on 1 case(s).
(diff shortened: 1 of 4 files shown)


jakirkham:
So you need to re-render after adding things to `yum_requirements.txt` in the feedstocks.

ccordoba12:
Thanks @jakirkham!

jakirkham:
Of course. 😄

ccordoba12:
Ok, I activated session management but for that I had to disable the dependency on `libuuid`.

@ceholden, how bad is that?

ceholden:
Hmm... I think I'd be back to the issue where compiling QGIS hits a linking error because GDAL and Qt are linking against two different versions of `libuuid`. Is there a way I can try this out before merging?

I'm wondering if we could drop `libuuid` from GDAL (via `gdal` -> `libdap4` -> `libuuid`). The [`libdap4` Github repo](https://github.com/OPENDAP/libdap4/) runs their tests on Travis CI and it looks like they get [`libuuid` from `apt-get`](https://github.com/OPENDAP/libdap4/blob/cd38361439b511aadbf4f5500e6127fccf41f71e/.travis.yml#L49)

@ocefpaf Do you know more about this? The Docker image should have a copy of `libuuid.so` and it's a pretty fundamental package to doing anything on Linux. I'll have a go at submitting a PR to `libdap4` that removes the `libuuid` dependency. If it compiles and passes tests, it would clear up this issue of that comes up when GDAL and Qt intersect.

ocefpaf:
> @ocefpaf Do you know more about this?

I haven't got time to check this yet. But I think that having this `Qt` recipe here would fix it, no? As now both `Qt` and `libdap4` will be using the same `libuuid` (at least on Linux).

>  I'll have a go at submitting a PR to libdap4 that removes the libuuid dependency. If it compiles and passes tests, it would clear up this issue of that comes up when GDAL and Qt intersect.

Worth a shot! :+1:

ccordoba12:
@ocefpaf, should I wait for @ceholden to try to remove the libuuid dependency of GDAL before merging this one?

ocefpaf:
This fixes an important issue. So let's merge it. I'll help @cholden debug
this and if needed we build qt again. (I believe the problem is elsewhere
and not in qt.)

I am away from my laptop now. Feel free to merge it.

On Oct 14, 2016 8:39 PM, "Carlos Cordoba" <notifications@github.com> wrote:

@ocefpaf <https://github.com/ocefpaf>, should I wait for @ceholden
<https://github.com/ceholden> to try to remove the libuuid dependency of
GDAL before merging this one?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/conda-forge/qt-feedstock/pull/9#issuecomment-253944702>,
or mute the thread

ccordoba12:
Ok, let's go with it :-)

ccordoba12:
I'll wait until packages are up in anaconda.org to rebuild pyqt :-)

PR #12, opened 2016-12-09
Win

ocefpaf:
@ccordoba12 I tried to build this in my old WindowsXP machine and failed miserably. I am sending this PR only to start a discussion on what needs to be done to get this to build locally so we can perform the manual upload later.

@gillins do you have access to all `vc`s? Can you try this recipe?


recipe/meta.yaml
```diff
@@ -31,15 +31,13 @@ source:
...
-  skip: true  # [win or osx]
   features:
     - vc9  # [win and py27]
     - vc10  # [win and py34]
     - vc14  # [win and py35]
 
 requirements:
   build:
-    # HACK: this is here to get Windows features (vc??) to work.
     - python  # [win]
```
ccordoba12: I this still needed?

gillins:
Yes I have all `vc`s installed. I remember doing this wasn't easy, but I got there in the end. Can't remember all the exact steps. I might be able to help if you are having problems doing this - although time to for you to upgrade from XP I think!

Just compiled Python 3.5 (VC 2015) fine using the recipe. Took 3 hours. Will run the other `vc`s and get back to you.

Great to see some progress on this BTW.

ccordoba12:
I think the recipe still needs some tweaks before being ready to be built, i.e. removing the Python build dep and adding the vc packages.

gillins:
Yes I agree, but does seem to compile fine for me on all VC versions. Let me know if I can help get these packages ready.

recipe/meta.yaml
ocefpaf: `conda-smithy` needs that to get the matrix correctly. But we don't need that for recipes that will not use conda-forge machinery. (And we are looking into updating `conda-smithy` so that is no longer needed.)

ocefpaf:
>  although time to for you to upgrade from XP I think!

Indeed I should... But that was the last time I bought a Windows and I am not looking forward to do that again :smile:

> I think the recipe still needs some tweaks before being ready to be built, i.e. removing the Python build dep and adding the vc packages.

Adding the `vc` packages yes (just did that btw), but we are not ready to remove the python dependency yet.

>  I remember doing this wasn't easy

Sorry to ask something so hard. I know `qt` is quite a beast to compile. However, if you succeed we will be able to upload those packages manually and end the jpeg 8 vs 9 grief. (At lest on qt4 land.)

PS: @gillins store those binaries if you succeed and ping me privately over email or gitter later. I will arrange some tests with our stack and I will upload them to the channel.

ocefpaf:
@conda-forge/qt we are actually getting a few successful Windows builds here! I might merge this and get at least a few of the matrix items uploaded.

ocefpaf:
@msarahan can you take a look at the `OS X` failures? It seems to be related with the long prefixes.

ccordoba12:
Yep, please merge to get those builds uploaded :-)

ocefpaf:
If re-starting does not get everything build in less the 1-hour limit we will use the local builds.

PR #13, opened 2016-12-10
No conda-build 2 for now

ocefpaf:
[skip appveyor]

I removed `conda-build 2` on Linux and `OS X` to test if the long prefix is indeed the problem here.


ocefpaf:
@ccordoba12 will tackle the `OS X` builds. Closing this for now.

PR #18, opened 2016-12-26
Try Python 3/6

ocefpaf:


recipe/meta.yaml
```diff
@@ -41,7 +41,7 @@ build:
   features:
     - vc9  # [win and py27]
     - vc10  # [win and py34]
-    - vc14  # [win and py35]
+    - vc14  # [win and py>35]
```
ocefpaf: Is `vc14` the correct option for Python 3.6?

ccordoba12:
Please skip `osx` to avoid wasting valuable build time on it. I mean, we know it'll always time out :-)

recipe/meta.yaml
ccordoba12: I believe so, yes :-)

ocefpaf:
> Please skip osx to avoid wasting valuable build time on it. I mean, we know it'll always time out :-)

Agreed. I will leave that there for now though because people are working on it and I do not want them to deal with conflicts.

PR #20, opened 2016-12-29
Conda build 2

ocefpaf:
Note to self: a quick test on Travis-CI with `j 2` still resulted in a test timeout.

```
cd "/Users/travis/miniconda3/conda-bld/recipe_1483010385751/work/qt-everywhere-opensource-src-4.8.7/./tools/linguist/lconvert"

/Applications/Xcode.app/Contents/Developer/usr/bin/make Makefile

make[4]: `Makefile' is up to date.

compiling ../shared/numerus.cpp

compiling ../shared/translator.cpp

compiling ../shared/translatormessage.cpp

compiling ../shared/qm.cpp



The job exceeded the maximum time limit for jobs, and has been terminated.
```

PR #21, opened 2016-12-30
patch for qtwebkit, gcc release 6*

neok-m4700:
when building with gcc in the release 6* series on linux, then enabling -webkit is not taken into account (reset to no) because the configure script does not support gcc > 5.x.


ccordoba12:
Thanks for the patch. Please wait until PR #20 is merged, then you have to rebase this one and bump the build number.

ocefpaf:
@neok-m4700 take a look at http://conda-forge.github.io/docs/ciservices.html#circleci-linux to see how you can enable CircleCI (Linux) testing here.

neok-m4700:
Hi have some timeout on Travis-CI, do you known where it does come from ?

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

ccordoba12:
> do you known where it does come from ?

Yes, but you don't need to worry about that error. We use Travis to compile macOS packages, but Qt takes more time to compile than the one provided by Travis.

neok-m4700:
Thanks,

By the way I also had to add CONFIG += uitools to projects.pro in order to build libQtUiTools.a
to projects.pro, this is needed when building pyside (python bindings to qt). Should I add this to the build.sh script ?

ocefpaf:
@neok-m4700 this looks good to go. Please rebase with latest master (to get rid of the OS X failing build) and bump the build number in the `meta.yaml` file (to get a new version with the patch uploaded).

Thanks for the patch!

ocefpaf:
The Windows failure is due to a bad AppVeyor update. I am merging this and fixing the lint in another PR. Thanks @neok-m4700!

PR #23, opened 2017-01-16
set $QMAKESPEC on OSX using an activate script

gillins:
This fixes the problem on OSX where building the test `hello.pro` with `qmake` fails since the paths for the default spec are incorrect because they got copied from the `unsupported` folder. This PR tells `qmake` what spec to use via the $QMAKESPEC environment variable which seems the documented way to set this. This should also mean that other packages using `qmake` get the correct spec.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -75,6 +75,16 @@ if [ $(uname) == Darwin ]; then
 
     make -j $CPU_COUNT
     make install
+
+    # copy activation scripts for OSX - set $QMAKESPEC
+    # to unsupported/macx-clang-libc++ otherwise there are problems with qmake
+    ACTIVATE_DIR=$PREFIX/etc/conda/activate.d
+    DEACTIVATE_DIR=$PREFIX/etc/conda/deactivate.d
+    mkdir -p $ACTIVATE_DIR
+    mkdir -p $DEACTIVATE_DIR
+
+    cp $RECIPE_DIR/scripts/activate.sh $ACTIVATE_DIR/qt-activate.sh
+    cp $RECIPE_DIR/scripts/deactivate.sh $DEACTIVATE_DIR/qt-deactivate.sh
 fi
 
 
--- a/recipe/scripts/activate.sh
+++ b/recipe/scripts/activate.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 
 if [[ -n "$QMAKESPEC" ]]; then
     export _CONDA_SET_QMAKESPEC=$QMAKESPEC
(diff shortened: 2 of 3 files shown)


ccordoba12:
+1 from me since this work for you locally.

@jakirkham?

jakirkham:
I'd feel better if we could avoid activate scripts. However, if setting an environment variable is the only way, activate scripts are definitely the way to go.

jakirkham:
Any other thoughts on this? @msarahan? @mingwandroid? Others?

ccordoba12:
Most probably this is needed because you're using an unsupported qmake spec. I say it because we have the same test internally and it passes just fine.

mingwandroid:
Can you not use `qmake -spec ${CONDA_PREFIX}/mkspecs/unsupported/macx-clang-libc++ hello.pro` instead?

Having said that, I have no issue with {de,}activate scripts. Fact is, some software needs env. vars set and sourcing a script is the only way to do that.

I think I would like for our Qt packages to provide debug libraries too before I was totally enthusiastic about adding it here, but I'm still OK with it.

Though I think there are some enhancements that need to be made so that installation into an already activated env. runs the activate script and removal from the same runs the deactivate script, I keep meaning to file an enhancement request about that. cc @kalefranz.

For some compiler packages, I've started to develop a generalized environment variable {re,}setting function for {de,}activate scipts that supports a single level of backup, if you were interested: https://github.com/mingwandroid/conda-recipes/blob/master/cross-compilers/compiler_linux-64_linux-aarch64/activate.sh (though it's overkill for your needs here).

gillins:
> Can you not use qmake -spec ${CONDA_PREFIX}/mkspecs/unsupported/macx-clang-libc++ hello.pro

Yes that also works, but I was after a solution that would work for other packages that use `qmake` without them having to do something different for OSX and have the spec hard-coded into their `build.sh`.

mingwandroid:
Yup understood, I'd give it my wholehearted support if we made debug libraries, but you have my support anyway (FWIW!)

jakirkham:
Thanks for taking a look @mingwandroid. 😄

kalefranz:
> Though I think there are some enhancements that need to be made so that installation into an already activated env. runs the activate script and removal from the same runs the deactivate script, I keep meaning to file an enhancement request about that.

That can't happen until `conda` is a shell function first, and only secondarily an executable.  It's high on my wish list.

gillins:
@jakirkham do you get a successful build with this recipe?

jakirkham:
Sorry have not had a chance to try.

gillins:
Is this good to be merged?

ocefpaf:
Yep. This waited long enough and it is working for you locally.

PR #24, opened 2017-01-17
Not ready yet Qt5

ocefpaf:
This is an attempt to get @mingwandroid's recipe from `qttesting` in `conda-forge`. I got this to build on Linux locally, let us see how it goes here.


recipe/meta.yaml
```diff
@@ -1,54 +1,23 @@
...
-
-    # Disable warning about Windows 10 not being a supported platform by Qt
-    - disable_win10_warning.patch  # [win]
-
-    # On Windows, for VS 2015 (Python 3.5+), rather drastic modifications are
-    # necessary because Microsoft finally got around to defining some standard
-    # functions. There's the vs_2015_int.patch file that does this. The changes in this patch were
-    # found in forums, and originated at (Author appears to be Peter Zhigalov):
-    # https://fami.codefreak.ru/gitlab/peter/qt4/commit/45e8f4eef3923e03c6939d0c17170980685857ef.diff
-    - vs2015_int_fix.patch  # [win and py >= 35]
```
ocefpaf: I did not check if any of the old patches are needed yet. I just removed them all and added the new ones.

recipe/meta.yaml
```diff
@@ -57,45 +26,57 @@ build:
...
   build:
     - python  # [win]
-    - freetype 2.6.*  # [unix]
-    - fontconfig 2.11.*  # [linux]
-    - freeglut  # [linux]
+    - perl
+    - freetype 2.7|2.7.*
+    - fontconfig 2.12.*
+    #- pango  # [linux]
+    #- freeglut  # [linux]
```
ocefpaf: All the commented out dependencies are those that are not present in @mingwandroid's version. We need to check if we really do not need them. I did notice that his `yum_requirements.txt` installs some of these.

gillins:
Should this be on a separate branch? We might need to build both Qt4 and Qt5 for a while until all software is ported over. Apologies if it is that already...
Thanks for pushing on with this @ocefpaf

ccordoba12:
@gillins is right, this needs to be in a different branch in case there's need to make new packages for Qt4.

ccordoba12:
But great work at making this pass!

recipe/meta.yaml
ccordoba12: Qt4 and Qt5 have different sets of dependencies, so several of the old ones don't apply for Qt5.

recipe/meta.yaml
ccordoba12: Old patches don't apply for Qt5.

jakirkham:
So typically `master` is the latest version of the package and we add branches to the main feedstock for old versions still in maintenance (e.g. Qt4). As all branches on the main feedstock are designed to deploy, this works flawlessly.

That being said, it would be nice if we could finish up PR ( https://github.com/conda-forge/qt-feedstock/pull/23 ) before this goes in `master`. Though it is not strictly necessary.

ccordoba12:
So we need to create a `qt4` branch for the old package?

recipe/meta.yaml
ocefpaf: We may need to adapt some of the old Windows patches. I won't look into that now though.

ocefpaf:
> Should this be on a separate branch? We might need to build both Qt4 and Qt5 for a while until all software is ported over. Apologies if it is that already...

I prefer to put `Qt4` in a branch and use master for `Qt5`, this is usually what we do. Do you guys agree?

PS: I will wait for #23 to be merged to do so.

ci_support/run_docker_build.sh
```diff
@@ -49,13 +49,22 @@ source run_conda_forge_build_setup
 yum install -y libX11-devel libXt-devel libXext-devel chrpath libXrender-devel gtk2-devel dbus-devel libSM-devel libICE-devel
 
 
-# Install conda-build 2.x.
-conda install --yes --quiet conda-build=2
-conda info
+# Embarking on 3 case(s).
+    set -x
+    export CONDA_PERL=5.20.3.1
```
ocefpaf: That is bad... We don't need this here.

jakirkham: Why not? Perl is set in `build`.

Admittedly I'm only able to see this part of the diff now. So if you are getting multiple Perls that is an issue. ( https://github.com/conda-forge/conda-smithy/issues/415 ) Would appreciate your thoughts on how we solve that.

recipe/meta.yaml
```diff
@@ -49,15 +43,10 @@ requirements:
...
     - fontconfig 2.12.*
-    #- freeglut  # [linux]
     - openssl 1.0.*
     - jpeg 9*
-    #- libtiff 4.0.*
     - zlib 1.2.*
     - libpng >=1.6.28,<1.7
-    #- jom  # [win]
-    ## This is here because conda is not pulling libxml2 from conda-forge
-    #- libxml2 2.9.3
```
ocefpaf: I removed all the dependencies that did not match @mingwandroid's recipe from `qttesting`. If @ccordoba12 and @mingwandroid could take a look and give a :+1: it would be awesome :wink:

ocefpaf:
> So we need to create a qt4 branch for the old package?

Done! @gillins I would like to merge #23 before this one, but if that does not happen would you mind re-directing #23 to the `qt4` branch?

recipe/build.sh
```diff
@@ -5,134 +5,104 @@
...
-                -no-harfbuzz \
-                -no-xinput2 \
-                -no-xcb-xlib \
-                -no-libudev \
-                -no-egl \
-                -no-openssl
-
-    DYLD_FALLBACK_LIBRARY_PATH=$PREFIX/lib make -j $MAKE_JOBS || exit 1
-    make install
+./configure -prefix $PREFIX \
```
ocefpaf: I tried to aggregate the common parts of Linux and OS X here to reduce the maintenance burden. Not sure if that was a good idea. What do you thing @mingwandroid?

ccordoba12: I really find it more confusing and prefer to leave it as it was (I was the one who originally created this file and wanted to clearly see what options are used in each OS separately).

Building Qt is a really complex process to add more complexity to it just save 40 lines of code.

recipe/meta.yaml
ocefpaf: Well... in a way this is clearer as you can see the common options in one place and the different options in another. But I am +0 to this one. I just find it easier to maintain this way as it helps to keep the options in sync.

recipe/build.sh
ccordoba12: If you're +0, I'm +1 on reverting for now. I'll factor the common options in a less confusing way this weekend.

ocefpaf: I actually meant +0.5 (typing on cellphone is hard!) I do prefer this as I find it quite clearer, but I am not married to the idea. If you feel strongly about it I will revert as soon as I am back home (I am in the road now).

However, I am curious to what you find less confusing did you see the `build.sh` alone? With the diff? It is quite easy to find what is common, OS X, or Linux.

gillins:
>  would you mind re-directing #23 to the qt4 branch?

Done. Let me know if/when you want me to look at the Windows build for Qt5. I think we are still waiting for a patch to use `webkit` for `py<35`?

ccordoba12:
> I think we are still waiting for a patch to use webkit for py<35?

Those patches were created by @mingwandroid and they are part of our package too.

mingwandroid:
IMHO you should use Webkit on all Windows versions.

ccordoba12:
> IMHO you should use Webkit on all Windows versions.

Agreed. Just to have the same Web backend for all VS versions.

ocefpaf:
> Done. Let me know if/when you want me to look at the Windows build for Qt5. I think we are still waiting for a patch to use `webkit` for `py<35`?

Thanks @gillins!

@ccordoba12 and @mingwandroid let's merge this and my next PR will add the Windows stuff.

ccordoba12:
Great job @ocefpaf!

ocefpaf:
> Great job @ocefpaf!

Only possible with great reviewers and my amazing copy-n-paste abilities from your's and @mingwandroid's work :stuck_out_tongue_winking_eye:

ccordoba12:
Great job at moving the ball then ;-) I'm so swamped that I'd have never done this!

ocefpaf:
> Great job at moving the ball 

That credit I can take :smile: 

>  I'm so swamped that I'd have never done this!

But I will bug you in `pyqt` soon :wink:

ccordoba12:
> But I will bug you in pyqt soon

No prob, but PyQt is very easy compared to Qt ;-)

hadim:
Nice work here guys !

Do you plan to add a 5.7 (and 5.8?) build ?

ccordoba12:
Qt 5.7+ needs a C++ 11 capable compiler, something that leaves Python 2.7 on Windows out of the picture.

So I guess that's a no, but I can't answer for all maintainers.

hadim:
Ok. But do you think it could be compiled on Python > 3 (for Linux first and then OSX and Windows) ?

That would be good to have those 5.7 and 5.8 versions here.

Wheels are provided for PyQt5 on pypi (only 5.7 at the moment). So there is no rush as it's a good alternative. It would just more convenient to have everything on conda-forge. But I know how difficult it is to compile Qt :-)

ccordoba12:
> Ok. But do you think it could be compiled on Python > 3

You could give it try with the current recipe, but I think @ocefpaf wants PyQt 5.6 packages first.

PR #27, opened 2017-02-05
Windows Qt5 builds

gillins:
This is an attempt to get the Qt5 builds going. Seems that `qtwebengine` isn't built by default so we don't have to do anything fancy there.

I'm skipping py27 on Windows until https://github.com/conda-forge/icu-feedstock/pull/16 is merged.

Currently I'm using `-opengl desktop` not sure if this is correct or not.


gillins:
Can I stop it doing a separate build for each version of Perl??

ccordoba12:
I think you need to pin Perl somehow. @jakirkham?

jakirkham:
Yep, you need to pin `perl`. Pick a version that works and pin to it exactly for now.

TBH having a `perl` matrix is pretty silly. However the behavior traces back to `conda-build`. We should have a frank discussion about how to handle `perl` in a smarter way.

xref: https://github.com/conda-forge/conda-smithy/issues/415

recipe/bld.bat
```diff
@@ -1,55 +1,51 @@
...
+     -system-libjpeg ^
+     -system-libpng ^
+     -system-zlib ^
+     -mp
+if errorlevel 1 exit /b 1
+
+:: re-enable echoing which is disabled by configure
+echo on
+     
+:: Note - webengine only built when you ask (nmake module-webengine) - so we can skip it easily.
```
ccordoba12: This is not good. We need to compile `webkit` or `webengine` (`webkit` is easier) or Spyder will break.

ccordoba12: Besides, `webengine` doesn't work with VS 2008, so it can't used with Python 2.7

recipe/bld.bat
```diff
@@ -1,55 +1,51 @@
...
+     -opensource ^
+     -openssl ^
+     -platform win32-msvc%VS_YEAR% ^
+     -release ^
+     -shared ^
+     -qt-freetype ^
+     -system-libjpeg ^
+     -system-libpng ^
+     -system-zlib ^
+     -mp
```
ccordoba12: What is this option used for?

gillins: It passes the [/mp](https://msdn.microsoft.com/en-us/library/bb385193.aspx) option to `cl`. I found building with `jom` just randomly failed. Building without `jom` but with `/mp` seems to keep the build time down to about 1 hour.

recipe/bld.bat
gillins: Yes I made this note so that it is clear that `webengine` isn't build by default (which we can't use as you say). I've just noticed that `webkit` wasn't being built locally (the sources are included in the Qt tree) so I am just about to follow the instructions [here](https://pempek.net/articles/2015/10/18/compiling-qt-5-5-1-with-visual-studio-2015/) and add `gnuwin32\bin` to the `PATH` to see if that fixes it (`ruby` should already be on AppVeyor). Any other tips appreciated...

gillins:
@ccordoba12 still doesn't seem to be building `webkit`. Any suggestions? I can't even see where the tests are done to decide what to build. 

Any thoughts on what to set for OpenGL?

ccordoba12:
> @ccordoba12 still doesn't seem to be building webkit

Webkit is not part of the distributed Qt5 tarballs anymore in 5.6 or after. You need to download it separately and move it to the build directory by yourself. This is how we do it in Continuum:

https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L46-L54

> Any thoughts on what to set for OpenGL?

I think using `desktop` is fine for now. Else you'd need to have the DirectX SDK:

https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L18-L24

However, @mingwandroid knows better about this.

gillins:
Uh Oh. Anyone have an idea as to why AppVeyor doesn't want to run this?

ocefpaf:
> Uh Oh. Anyone have an idea as to why AppVeyor doesn't want to run this?

Let's try closing and opening it again.

gillins:
Thanks @ocefpaf ! I must remember that trick.

mingwandroid:
@gillins, is there any reason to deviate https://github.com/gillins/qt-feedstock/blob/1bd83b8614d1fadc9e323cda155e5f0557f5e09e/recipe/bld.bat from https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat so much?

My opinion is that it makes interoperability more difficult without benefit. Sure I get that there needs to be changes to make it work in conda-forge, but mostly, those can be achieved with `::` at the start of lines you don't need.

ocefpaf:
@gillins the AppVeyor issue is an API rate limit that fails without any warning. We need to wait little bit and do a close/open or send a new commit.

@mingwandroid I did not see the changes but I agree with you. Even it is best to try to keep the recipes in sync. However, if there are any important changes that @gillins made, maybe we could take a look and absorb them into the original recipe?

PS: I believe that the Python 2,7 error, `error C1083: Cannot open include file: 'stdint.h': No such file or directory`, can be fixed with the `msinttypes` packages,

jakirkham:
> Uh Oh. Anyone have an idea as to why AppVeyor doesn't want to run this?

xref: https://github.com/conda-forge/conda-forge.github.io/issues/314

gillins:
@mingwandroid After [you suggested](https://github.com/conda-forge/qt-feedstock/pull/24#issuecomment-274098837) just using webkit for all versions I guess I saw the conda-forge one being different since it would be simpler. There was a bunch of stuff I didn't understand in the anaconda one like the [custom Py27 environment](https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L91-L100), [running conda-forge code manually](https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L16), [webengine defs](https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L106-L126) and [rebuilding configure.exe](https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L151-L189). I wondered how much of this was still needed (and would just confuse us down the track) so I just started with the basics and worked up.

However I don't have any strong feelings. I won't be able to do any more work on this for the next ~3 weeks anyway so feel free to open a new PR with your `bld.bat`.

Thanks @ocefpaf seems we are hitting the 90 minute limit on AppVeyor now... I suppose this won't be increased any time soon?

mingwandroid:
I didn't mean to denigrate your hard work @gillins, so apologies if it seemed that way. I hoped the comments in the blocks you linked to (https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L91-L100 and https://github.com/ContinuumIO/anaconda-recipes/blob/master/qt/bld.bat#L91-L100) explain in detail the code that proceeds them, if not I'm happy to try to give more details.

For the webengine stuff, I spent a while working on that, but ultimately VS < 2015 wasn't up to the task and we favoured consistency on Windows rather than splitting the backends. I wanted to leave the webengine stuff in there in-case we (or conda-forge) decided to split the web backends later. I don't much harm in having skipped logic in there or commenting it out for future reference as it can save duplication of effort later. I guess you go for a more minimal approach though.

I expect I'll not manage to find time for any PRs for this.

gillins:
@mingwandroid definitely see your point, happy to work with you on merging code when I get back if you haven't had time to do this.

ccordoba12:
Too bad this is timing out. We'd need 2 hours, maybe 2 hours and a half for this.

ocefpaf:
I will make another request to AppVeyor but I am not sure we can get more than what we already.
But we did manual uploads for `Qt4` on windows before and we can do it again as long as people are OK with this recipe.

gillins:
@mingwandroid just having another look at this. Questions I have are:
- Why do you need to have an explicit call to `obvci_appveyor_python_build_env.cmd`?
- Is the DirectX SDK stuff needed? Angle?
- Why do you need a custom Python 2.7 environment?
- Do we need to keep the rebuilding of `configure.exe`? The mentioned patch no longer seems to be applied.

mingwandroid:
Details may be slightly hazy but ..

> Why do you need to have an explicit call to obvci_appveyor_python_build_env.cmd?

conda-forge won't need that. I wanted to borrow some of the Win7 SDK setup for the non-conda-forge build.

> Is the DirectX SDK stuff needed? Angle?

Mostly, but *possibly* for more, audio, input?

> Why do you need a custom Python 2.7 environment?

Webkit needs Python 2.7 for some codegen purposes.

> Do we need to keep the rebuilding of configure.exe? The mentioned patch no longer seems to be applied

No, it's skipped, but thinking about it, since we try to build software from source, I'd prefer to unconditionally always rebuild configure.exe.

gillins:
Do we still need `tst_compiler.cpp`? This didn't make it over into conda-forge.

mingwandroid:
I don't know anything about that file. @ccordoba12?

ccordoba12:
I have no idea about it, but it seems you added it @mingwandroid because it's used in `bld.bat` and I didn't touch the Windows stuff for Qt :-)

gillins:
@mingwandroid how does this look? I have moved the `tst_compiler.cpp` part into the `confgure.exe` rebuilding section (which is where it probably belongs) which isn't currently run for now.

gillins:
Is this good to merge?

ccordoba12:
@gillins, do you know if QtWebkit gets built correctly?

gillins:
I see it being built - yes. But I haven't actually done anything with it. Should there be a test for this somewhere?

ccordoba12:
It's not easy to create pure Qt tests, much less run them on Windows. However, I'd really like a confirmation that QtWebkit is built (else Spyder won't work with these packages).

So I don't know what to do....

gillins:
I can confirm that `Qt5WebKit.dll` and `Qt5WebKit.lib` are created under `vc14`. I'm just running `vc9` to check that.

ocefpaf:
> It's not easy to create pure Qt tests, much less run them on Windows. However, I'd really like a confirmation that QtWebkit is built (else Spyder won't work with these packages).

> So I don't know what to do....

There is not much we can do besides uploading them, testing against Spyder, and removing if things are broken. Maintaining `test` channels is an alternative but an extra burden. I'd rather just do this over a weekend.

recipe/meta.yaml
```diff
@@ -26,31 +32,34 @@ build:
...
   build:
     - python  # [win]
+    - 7za  # [win]
+    - curl  # [win]
+    - msinttypes  # [win and py27]
     - perl 5.22.2.1
-    - freetype 2.7|2.7.*
-    - fontconfig 2.12.*
+    - freetype 2.7|2.7.*  # [unix]
+    - fontconfig 2.12.*  # [linux]
```
jakirkham: Why does `fontconfig` get a Linux instead of a Unix selector?

recipe/meta.yaml
```diff
@@ -26,31 +32,34 @@ build:
...
     - openssl 1.0.*
     - jpeg 9*
     - zlib 1.2.*
     - libpng >=1.6.28,<1.7
-    - gst-plugins-base
+    - gst-plugins-base  # [linux]
     - icu 58.*
-    - libxcb
-    - dbus
+    - libxcb  # [linux]
```
jakirkham: Given we have this on all platforms, we probably could drop the selector, but am not sure if there is a benefit in doing it. Maybe someone else can weigh in.

cc @pkgw

recipe/meta.yaml
ccordoba12: Because it's not needed in macOS.

recipe/meta.yaml
ccordoba12: Please leave things as they are, `libxcb` is only needed on Linux. The list of requirements to build Qt5 on each platform is listed [here](http://doc.qt.io/qt-5/build-sources.html) and `libxcb` is only mentioned on the Linux seciton.

Qt5 is a big and complex enough package to start meddling with its dependencies just to gain a little bit of consistency.

ccordoba12:
> I can confirm that Qt5WebKit.dll and Qt5WebKit.lib are created under vc14

@gillins, could you add a test that verifies that `Qt5WebKit.dll` is present in the built package? You can take inspiration for it from the Boost feedstock.

recipe/meta.yaml
jakirkham: Why not? Is it just using the system copy instead? That would be a problem as we use `fontconfig` as a dependency of many other packages on OS X.

recipe/meta.yaml
jakirkham: Before @pkgw started packaging X11 on all platforms I would have agreed we can just ignore X11 anywhere but Linux. However, it has since become clear to me that people are very interested in a working X11 ecosystem on all platforms. So I think it is fair for them to have the opportunity to weigh in. Personally I have no stake in this.

mingwandroid:
Just because dependencies can be used by a project on one OS it does not follow the same ones can be used without significant effort on other OSes. In my opinion we should defer to @ccordoba on Qt on macOS since he knows a lot about it.

Implementing support for these would require a huge amount of effort. What support you could achieve for libxcb is debatable and would lead to a non native windowing, decorations and general experience for all Qt applications and while fontconfig is probably less effort it would lead to non-native font handling .. though this one is somewhat interesting to me for reasons of font selection and consistency across platforms. I am not putting my hat in the ring to do this though.

These changes would require maintainence going forwards as the Qt Project would likely have little interest in this.

gillins:
`Qt5WebKit.dll` is also created under `vc9` and I've added a check. Are we good to merge this?

ccordoba12:
+1 from me to merge too.

gillins:
I'll build some packages and email @ocefpaf Not sure what the long term plan is for getting this working on the CI's...

ocefpaf:
Thanks for that!

I'm asking AppVeyor again but I don't think they can give as more than we
already have :-(

recipe/meta.yaml
pkgw: Working X11 ecosystem, yes, but it's very reasonable for graphical toolkits to use system-native backends as appropriate. I don't see any need to add the dep on non-Linux here.

PR #30, opened 2017-03-12
Add xorg-x11-server-Xvfb for X server test...

ocefpaf:
... and update the ICU pin so we can try it with the rest of the package stack later.

Note that we cannot move to `Qt5` yet since we still don't have `Windows` and `OS X` build. Both should break the time limit and volunteers will be needed to build locally and upload the packages (unless we reach an agreement with regards the pinnings with `defaults` :grimacing:).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
 BSD 3-clause license
-Copyright (c) conda-forge
+Copyright (c) 2015-2017, conda-forge
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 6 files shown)


ccordoba12:
@ocefpaf, you should add a test that involves the graphical part of Qt to verify that this is working.

ocefpaf:
> @ocefpaf, you should add a test that involves the graphical part of Qt to verify that this is working.

That is the goal but I don't know how :grimacing: 

I did test `pyqt` with this locally and it works on Linux. That is my end goal here. But if you can suggest a quick test I would gladly add it in this PR.

ccordoba12:
I'd be happy to help but my knowledge of pure Qt is also very limited ;-)

> I did test pyqt with this locally and it works on Linux

So my suggestion is to add `xorg-x11-server-Xvfb` to the PyQt feedstock instead of here.

ocefpaf:
> So my suggestion is to add xorg-x11-server-Xvfb to the PyQt feedstock instead of here.

I already add it there too but my goal adding it here is so we can also perform a similar test here. I am researching example on how to do it. Let me know if you have any suggestions.

Issue #34, opened 2017-03-23
undefined symbol in phonon for osx


jschueller:
Since qt(4) is in conda-forge, the build of pyside fails for osx since tests attempt to load phonon with missing symbols (https://s3.amazonaws.com/archive.travis-ci.org/jobs/214258733/log.txt):
```
from PySide import phonon

ImportError: dlopen(/Users/travis/miniconda3/conda-bld/pyside_1490279798328/work/PySide-1.2.4/sources/pyside/build/PySide/phonon.so, 2): Symbol not found: __ZN6Phonon22ObjectDescriptionModelILNS_21ObjectDescriptionTypeE0EE11qt_metacastEPKc

  Referenced from: /Users/travis/miniconda3/conda-bld/pyside_1490279798328/work/PySide-1.2.4/sources/pyside/build/PySide/phonon.so

  Expected in: flat namespace

 in /Users/travis/miniconda3/conda-bld/pyside_1490279798328/work/PySide-1.2.4/sources/pyside/build/PySide/phonon.so
```

```
$ c++filt _ZN6Phonon22ObjectDescriptionModelILNS_21ObjectDescriptionTypeE0EE11qt_metacastEPKc
Phonon::ObjectDescriptionModel<(Phonon::ObjectDescriptionType)0>::qt_metacast(char const*)
```

Sounds a bit like https://bugreports.qt.io/browse/QTBUG-37209

jschueller:
you included it recently @jakirkham, why is it marked test ?

jakirkham:
xref: https://github.com/conda-forge/qt-feedstock/issues/15#issuecomment-287202192

Though we should definitely test it with `pyqt` and `pyside`.

cc @gillins

gillins:
@jschueller have you tried recompiling Qt with the patch you linked to? Does it resolve the problem? If so might be worth putting in a PR so we can get @jakirkham to rebuild.

Have you tried `pyqt`?

jschueller:
the problem is when linking pyside.phonon, not qt. I did not try pyqt.

jakirkham:
Would be good to try PyQt when you have a chance as well.

jschueller:
I tried too: https://github.com/conda-forge/pyqt-feedstock/issues/19

jakirkham:
Ah, but we've only built Qt4 thus far. That doesn't work with PyQt5, correct?

jschueller:
[...]

PR #37, opened 2017-04-11
Fix phonon linking for Qt4 on OSX

gillins:
This PR fixes `pyside` linking to phonon. I was able to remove the `-E phonon` part from `pyside`'s `build.sh` and it build and tested fine.

See https://github.com/conda-forge/qt-feedstock/issues/34

I was unable to test `pyqt` due to https://github.com/conda-forge/pyqt-feedstock/issues/20

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -36,6 +36,9 @@ source:
     # found in forums, and originated at (Author appears to be Peter Zhigalov):
     # https://fami.codefreak.ru/gitlab/peter/qt4/commit/45e8f4eef3923e03c6939d0c17170980685857ef.diff
     - vs2015_int_fix.patch  # [win and py >= 35]
+    # see https://bugreports.qt.io/browse/QTBUG-37209
+    # fix phonon exports so they can be linked to
+    - phonon_osx_exports.patch  # [osx]
 
 build:
   number: 7
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -41,7 +41,8 @@ source:
     - phonon_osx_exports.patch  # [osx]
 
 build:
-  number: 7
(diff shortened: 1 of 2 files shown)


jakirkham:
@ccordoba12, could you please look at this one?

ccordoba12:
The patch seems fine (I mean, if @gillins says it work for him, I trust him).

This requires a bump in build number (although it'd be better to just bump the number for `osx`).

jakirkham:
So there is one, but it is under the `test` label. Probably better to bump in this case so we can distinguish the two.

gillins:
> Probably better to bump in this case so we can distinguish the two

done

gillins:
Tested `pyqt4` version 4 with https://github.com/conda-forge/pyqt-feedstock/pull/21. All ok.

ccordoba12:
Great work @gillins! Merging then :-)

jschueller:
great news!

jschueller:
shouldn't we bump the release number ?

recipe/meta.yaml
```diff
@@ -36,9 +36,13 @@ source:
...
     # https://fami.codefreak.ru/gitlab/peter/qt4/commit/45e8f4eef3923e03c6939d0c17170980685857ef.diff
     - vs2015_int_fix.patch  # [win and py >= 35]
+    # see https://bugreports.qt.io/browse/QTBUG-37209
+    # fix phonon exports so they can be linked to
+    - phonon_osx_exports.patch  # [osx]
 
 build:
-  number: 7
+  number: 7  # [linux or win]
+  number: 8  # [osx]
```
gillins: @jschueller It was bumped to 8 here

jschueller: ok, thanks

PR #38, opened 2017-04-12
Mac build requires Python 2

stuarteberg:
...since chromium doesn't build without it

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -34,6 +34,7 @@ build:
 requirements:
   build:
     - python  # [win]
+    - python 2.7*  # [osx]
     - 7za  # [win]
     - curl  # [win]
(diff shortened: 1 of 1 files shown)


recipe/meta.yaml
```diff
@@ -33,7 +33,7 @@ build:
 
 requirements:
   build:
-    - python  # [win]
+    - python 2.7* # [osx or win]
```
jakirkham: The Windows part has to do with `vc` features. So we can't pin it under the Windows selector. Though we could pin it under an `osx ` only selector.

stuarteberg: OK, thanks.  Now updated.

ccordoba12:
Thanks a lot @stuarteberg!

PR #40, opened 2017-04-19
Test that libQt5WebKit.so is built on Linux

ccordoba12:
WebKit was not built on Linux because there were several missing build deps. This PR fixes that.


ccordoba12:
@ocefpaf, @jakirkham, building on Linux also times out when we include QtWebKit. So I skipped it there too.

With my changes in this PR it should be possible to get a build with QtWebKit, but someone needs to do it locally. Do you have free time to do it?

I'm merging this because it seems ready. If more errors are encountered when building locally, new PRs will have to be submitted (hopefully not ;-)

jakirkham:
So I guess we can't build Qt5 on CIs anywhere then. 😞 Maybe we should spend some time discussing how best to handle building Qt5 long term then.

Unfortunately, I don't have time to build Qt. That said, this is an easy thing that anyone can do. Given one has `docker`, all one has to do is remove the `skip` locally, re-render, and run `ci_support/run_docker_build.sh`. This will build and produce a package in `build_artefacts/linux-64/`.

ccordoba12:
> Maybe we should spend some time discussing how best to handle building Qt5 long term then.

Having proxies is the best option in my opinion. Right now we have:

* @stuarteberg for macOS.
* @gillins for Windows.
* Me for Linux (I'll give it a go this weekend :-).

gillins:
I'm happy to build for Windows in the short term, but definitely vote for a finding better solution long term as it is quite a pain to build (especially if we start tweaking the recipe again)...

ccordoba12:
@ocefpaf, @jakirkham, I built Qt5 successfully with QtWebKit. It's on my channel now

https://anaconda.org/ccordoba12/qt/files

so please update it as soon as you can.

ccordoba12:
@jakirkham, could I ask you to move my package to conda-forge as soon as possible? We need it to trigger a recompilation in PyQt. That's needed to make Spyder work on Linux with Qt5.

ocefpaf:
> so please update it as soon as you can.

It would be nice to get a checksum for those. And it will take a while. For some reasons this service is very slow today on my end.

qt

ccordoba12:
> It would be nice to get a checksum for those.

Md5 for this file: a36f9b172d4dc71a51ff15d15015aa44

> For some reasons this service is very slow today on my end.

Yee, that's very slow! Please let me know when it's up :-)

ocefpaf:
Build `2` OS X should be in the channel now.

ccordoba12:
I guess you mean on Linux. Thanks a lot!

PR #42, opened 2017-04-24
Remove unneeded dependencies in macOS

### LINKED ISSUE #41: Limit runtime requirements (on osx)
When installing qt from conda-forge channel, the following packages are installed:
```
"conda-forge::ca-certificates-2017.1.23-0 1",
"conda-forge::expat-2.1.0-2 1",
"conda-forge::icu-58.1-1 1",
"conda-forge::jpeg-9b-0 1",
"conda-forge::zlib-1.2.11-0 1",
"conda-forge::dbus-1.10.10-3 1",
"conda-forge::libpng-1.6.28-0 1",
"conda-forge::openssl-1.0.2k-0 1",
"conda-forge::freetype-2.7-1 1",
"conda-forge::qt-5.6.2-1 1"
```

while qt from defaults channel only installs
```
"icu-54.1-0 1",
"zlib-1.2.8-3 1",
"qt-5.6.2-0 1"
```

The one I am most concerned about is dbus, as it kindly asks to add a Launch daemon which I do not want to do. Are all of them really needed or were they inherited from the linux build?

ccordoba12:
Fixes #41 

----

@stuarteberg, we need your help for a recompilation here. Thanks!

I'll leave this PR open in case you have problems and need to add more changes.

----

@ocefpaf, @jakirkham, please **stop** changing/touching dependencies for Qt5. It's clear you don't know when they are needed and when they aren't.

Examples for macOS:

* openssl is not needed since Qt 5.6. See https://github.com/Homebrew/legacy-homebrew/issues/42217#issuecomment-126379557
* You made the package to depend on system libraries (jpeg and libpng) but didn't change the configure options to make that happen.
* You made the package to depend on dbus when there's no need for that.

Examples for Linux:

* You didn't include bison, flex, gperf and ruby, even when they are listed in Continuum's recipe. That's the reason why QtWebKit was not built.

This only makes to waste my and the rest of maintainers time.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -113,11 +113,11 @@ if [ $(uname) == Darwin ]; then
                 -skip wayland \
                 -skip canvas3d \
                 -skip 3d \
+                -system-freetype \
+                -system-libjpeg \
+                -system-libpng \
                 -system-zlib \
                 -qt-pcre \
-                -qt-freetype \
-                -qt-libjpeg \
-                -qt-libpng \
                 -c++11 \
                 -no-framework \
                 -no-dbus \
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,8 +23,8 @@ source:
     - 0001-shobjidl-Fix-compile-guard-around-SHARDAPPIDINFOLINK.patch  # [win]
 
 build:
-  number: 1  # [osx or win]
-  number: 2  # [linux]
+  number: 1  # [win]
+  number: 2  # [unix]
   skip: True  # [osx or win or linux]
   detect_binary_files_with_prefix: true
   features:
@@ -43,7 +43,7 @@ requirements:
     - curl  # [win]
     - msinttypes  # [win and py27]
     - perl 5.22.2.1
-    - freetype 2.7|2.7.*  # [linux]
+    - freetype 2.7|2.7.*  # [unix]
     - fontconfig 2.12.*  # [linux]
     - openssl 1.0.*  # [not osx]
     - jpeg 9*
@@ -57,7 +57,7 @@ requirements:
     - vc 10  # [win and py34]
     - vc 14  # [win and py>=35]
   run:
-    - freetype 2.7|2.7.*  # [linux]
+    - freetype 2.7|2.7.*  # [unix]
     - fontconfig 2.12.*  # [linux]
     - openssl 1.0.*  # [not osx]
     - jpeg 9*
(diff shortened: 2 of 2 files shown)


jakirkham:
Sorry @ccordoba12, but I'm afraid I don't know what you mean. There are no changes from me in the git history of this feedstock's content let alone to its dependencies. Once had added a PR to properly restrict `gst-plugins-base` to Linux after issues building on Mac, which was closed out as it was handled by a different PR. As I'm also not listed as a maintainer of these feedstock and am generally trying to not be that involved in it, I'm unclear on what I have done wrong here.

As to the dependencies, FWICT by looking at the history in `git` either they have existed since the outset of this feedstock or at least before adding Qt5. One of them (`libpng`) merely saw a version constraint change and another (`dbus`) had a Unix selector applied when none existed previously. So it appears more likely to be a failure to remove them or properly restrict them than any additional add. These seem to have been added after a fair bit of discussion and consensus. All of this just to say, I don't think anyone has intentionally harmed this feedstock.

That all being said, I understand it can be frustrating when things don't work right in one's dependency stack. When it comes to something like Qt, which is crucial for many GUI applications, I can appreciate it takes a fair bit of work to get right and any problem with it is felt widely. However, if there is concern over what has been changed in the feedstock, a polite discussion with other maintainers should be sufficient for correcting course. Thanks for understanding.

ccordoba12:
> There are no changes from me in the git history of this feedstock's content let alone to its dependencies

Sorry to include you in this @jakirkham, I remember you wanted to change some deps from Linux only to Unix, so I thought you did that at the end.

> As to the dependencies, FWICT by looking at the history in git either they have existed since the outset of this feedstock or at least before adding Qt5

Then it was @ocefpaf's fault, when he first add the Qt5 recipe. Again, sorry for including you in this discussion.

> That all being said, I understand it can be frustrating when things don't work right in one's dependency stack

Yes, it's very frustrating for me to put more time helping conda-forge when Continuum's recipe has very clear dependencies and you don't use them as they are listed. At least you could ask me or Ray before changing them.

> a polite discussion with other maintainers should be sufficient for correcting course

I apologize if I was rude before. It's just that I have to spend a lot of time this week fixing things here, that shouldn't have been necessary if the recipe were copied exactly as Continuum created it.

Let's say that from now on every change in deps has to be approved by the maintainers.

ocefpaf:
> Then it was @ocefpaf's fault, when he first add the Qt5 recipe. Again, sorry for including you in this discussion.

I don't recall building it for OS X as I have zero OS X knowledge and no Mac hardware at all.

I **do** recall proposing a pinning scheme that would allow users to rely on conda-forge's qt/pyqt4 and ensure stack stability within conda-forge or qt/pyqt 5.* from defaults and ensure latest qt.

Anyways it is easy to point fingers and I don't really care for that.

Issue #50, opened 2017-08-24
installed Qt has no DB drivers


jschueller:
seems on windows qt is built without database plugins (on linux it seems ok)
I see we have sqlite packages, maybe it could be added as dependency ?

gillins:
More information here: http://doc.qt.io/qt-5/sql-driver.html

Not sure how it is working on Linux as documentation states default is no drivers. Feel free to submit a PR with how you think it should look (which drivers, plugins or builtins) and I can have a go at doing a new build...

gillins:
Fixed by #51

PR #51, opened 2017-08-25
Enable sql support

jschueller:
attempt to solve #50

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -49,6 +49,8 @@ if [ $(uname) == Linux ]; then
                 -system-libjpeg \
                 -system-libpng \
                 -system-zlib \
+                -system-sqlite \
+                -plugin-sql-sqlite \
                 -qt-pcre \
                 -qt-xcb \
                 -qt-xkbcommon \
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -160,6 +160,8 @@ call configure ^
      -system-libjpeg ^
      -system-libpng ^
      -system-zlib ^
+     -system-sqlite ^
+     -plugin-sql-sqlite ^
      -mp
 if errorlevel 1 exit /b 1
 
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,9 +23,9 @@ source:
     - 0001-shobjidl-Fix-compile-guard-around-SHARDAPPIDINFOLINK.patch  # [win]
 
 build:
-  number: 1  # [win]
-  number: 3  # [linux]
-  number: 4  # [osx]
+  number: 2  # [win]
+  number: 4  # [linux]
+  number: 5  # [osx]
   skip: True  # [osx or win or linux]
   detect_binary_files_with_prefix: true
   features:
@@ -52,6 +52,7 @@ requirements:
     - libpng >=1.6.28,<1.7
     - gst-plugins-base  # [linux]
     - icu 58.*
+    - sqlite 3.13.*
     - libxcb  # [linux]
     - dbus  # [linux]
     - vc 9  # [win and py27]
@@ -66,6 +67,7 @@ requirements:
     - libpng >=1.6.28,<1.7
     - gst-plugins-base  # [linux]
     - icu 58.*
+    - sqlite 3.13.*
     - libxcb  # [linux]
     - dbus  # [linux]
     - vc 9  # [win and py27]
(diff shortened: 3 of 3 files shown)


gillins:
You'll need to bump the build number also

jschueller:
ok

gillins:
I've uploaded packages for `win64` to the `gillins` channel for testing. The sha256 are:
qt-5.6.2-vc9_2.tar.bz2 830e4d46429cd608cef72654fb89836c867ef6b94f7689fe3d6b94552be80e3e
qt-5.6.2-vc14_2.tar.bz2 5fd1829b913a1d88e8f26ff597f558947198c692b50cbce38eb0fc114a69fb2a

Let me know how they go. I think you will have to add a test to make sure that `%LIBRARY_PREFIX%\plugins\sqldrivers\qsqlite.dll` exists (and presumeably `$PREFIX/plugins/sqldrivers/qsqlite.so` for Linux etc)

gillins:
Out of interest, do you have some software that needs this functionality?

Does anyone else have an opinion on whether this is a worthwhile feature to add?

mingwandroid:
I think Qt should be as fully featured as possible, esp. wrt these small bits since they're unlikely to have much impact on the (already large) size.

jschueller:
I also changed the installation dirs on linux to be on subdirectories instead of root.

jschueller:
@mingwandroid I dont see where it's used the second time: the other place CPU_COUNT is used directly

mingwandroid:
Ah that's true, well I'd rather it used twice rather than removed, debugging build issues in Qt is tricky so having access to disabling this at the top of the recipe is helpful but it's not very important either way.

jschueller:
CPU_COUNT can be overriden by the env, so it's even easier to only used that

mingwandroid:
No, `CPU_COUNT` is handled explicitly by `conda-build` and not something you can override in the environment.

recipe/build.sh
```diff
@@ -19,14 +19,12 @@ if [ $(uname) == Linux ]; then
...
     rm qtwebkit.tar
 
-    MAKE_JOBS=$CPU_COUNT
-
     ./configure -prefix $PREFIX \
                 -libdir $PREFIX/lib \
                 -bindir $PREFIX/bin \
                 -headerdir $PREFIX/include/qt \
-                -archdatadir $PREFIX \
-                -datadir $PREFIX \
```
ccordoba12: Why are you changing this? I think it will break every other lib that depends on Qt.

jschueller: I though it was cleaner to install plugins & stuff in a subdir, are you sure this will break something ? libs are still installed in libdir

jschueller:
@mingwandroid see https://github.com/conda/conda-build/pull/1149

recipe/build.sh
jschueller: I reverted in doubt

mingwandroid:
@jschueller, you know my preference here but if you prefer it this way I'm not going to argue any further.

jschueller:
maybe I can set it in anoter pr

gillins:
Is this PR good to merge? @jschueller did the win64 packages check out ok for you?

jschueller:
I think this is ready, but I could not test the win packages.

gillins:
Just building new Windows packages now....

gillins:
I've uploaded some new packages with the updated recipe to the 'gillins' channel. Here are the `sha256sum`s:
win-64/qt-5.6.2-vc14_2.tar.bz2: 7952a1cb8d97f0ab2f821948ca3f0f7aabeb302b815f3176886c9e8d4b7954db
 win-64/qt-5.6.2-vc9_2.tar.bz2: 10814dc15b58bbef85d8a7440cdbdc0c61993e4eef50a85c8b5a1e7bf69cb6ca
win-32/qt-5.6.2-vc14_2.tar.bz2: 4d90886afb9d7b83a055ee47098fed0ee52be071bfc3fc1985d2966e93e1aaed
win-32/qt-5.6.2-vc9_2.tar.bz2: fc7e4a519830a7ddd46e57a95c5112941d649be3b0b1880b0246cff25b76e30a

@ocefpaf Can you upload these please?
@stuarteberg Are you still building the OSX packages? If so, can you please update the `qt` build?

Who is in charge of doing the Linux builds now?

ocefpaf:
> I've uploaded some new packages with the updated recipe to the 'gillins' channel.

Thanks! I know how onerous those are to build.

> @ocefpaf Can you upload these please?

Done.

> Who is in charge of doing the Linux builds now?

On it... I'll experiment with the X11 packages we are shipping. That is something I wanted to do for a long time now.

PR #55, opened 2017-09-04
Remove useless MAKE_JOBS var

jschueller:
if necessary it can even be overriden via the env

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -19,8 +19,6 @@ if [ $(uname) == Linux ]; then
     patch -p0 < "${RECIPE_DIR}"/0005-qtwebkit-fix-TEXTREL-on-x86-changeset_172759.patch
     rm qtwebkit.tar
 
-    MAKE_JOBS=$CPU_COUNT
-
     ./configure -prefix $PREFIX \
                 -libdir $PREFIX/lib \
                 -bindir $PREFIX/bin \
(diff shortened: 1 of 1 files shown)


ocefpaf:
@mingwandroid are you OK with this solution? Should we merge it?

(Asking b/c I did not participated in the original discussion and I don't know decision made :smile:)

jschueller:
The thing is MAKE_JOBS was only used for Linux, not for osx, so at least it'll be consistent (and simpler). @mingwandroid concern was that you could not easily override this. The thing is you still can, via the env as CPU_COUNT passes trough conda-build.

Anyways this is a minor cleanup, if it doesn't get in I'm not going to insist.

mingwandroid:
IMHO the bug that needs to be fixed is that MAKE_JOBS should be used on macOS, but I too am not going to insist! Merging is fine by me.

jschueller:
MAKE_JOBS  does not pass through conda-build on the contrary to CPU_COUNT, so you may want to open an issue there

mingwandroid:
Not at all. You misunderstand the point here. It is to locally hack build.sh itself during development.

jschueller:
Yes, I don't understand.

mingwandroid:
Merging. There's no point continuing this discussion, you are being willfully obtuse. I explained the purpose behind this in plain English, clearly it is to suit my development methodology. That this isn't the same as yours is neither here nor there.

PR #60, opened 2017-11-09
Update pins

ocefpaf:
Linux build on my CircleCI instance. Let us see if we have enough CI time to build and upload on `master`.


recipe/meta.yaml
```diff
@@ -23,15 +22,13 @@ source:
     - 0001-shobjidl-Fix-compile-guard-around-SHARDAPPIDINFOLINK.patch  # [win]
 
 build:
-  number: 2  # [win]
-  number: 4  # [linux]
-  number: 5  # [osx]
```
ocefpaf: Multiple build number are elegant but error prone. In conda-forge we use a single one even if the other platforms did not change.

jschueller:
it seems that newest build 6 is not available with newer libpng pinnings and I suspect this is a problem with the newest matplotlib pinning

ocefpaf:
I check but what platform are you working on? OS X? Not that we only have Linux builds at the moment.

PR #61, opened 2017-12-07
update pins

ocefpaf:


mingwandroid:
Are you sure you want to pin to freetype 2.8.1 like this?

I am not saying you definitely do not mind you, but rebuilding Qt is a big job and having to do it on each patch release of freetype may not be possible.

ocefpaf:
What do you recommend? (That does not involves moving to conda-build 3 :stuck_out_tongue_winking_eye:)

mingwandroid:
2.8.*, it seems they did remove a load of symbols one time during a patch release last year but I don't think that it will be a common event.

ocefpaf:
> 2.8.*, it seems they did remove a load of symbols one time during a patch release last year but I don't think that it will be a common event.

I'll send that to the pinning scheme and then get to another round of re-builds. Thanks!

jakirkham:
Could we do something like `2.8.*,>=2.8.1` to avoid `2.8.0`?

mingwandroid:
I must be missing some details here, what's wrong with `2.8.0`?

My preference for pinning if you need to avoid `2.8.0` (and in general) is:

`>=2.8.1,<2.9.0a0`

This mirrors the pinning format that cb3 emits and handles corner cases like `pre` and `rc`.

jakirkham:
IDK, but I'm assuming there was some motivation for not using 2.8 (they don't use 0 for some reason). Unfortunately PR ( https://github.com/conda-forge/conda-forge.github.io/pull/494 ) which changed to 2.8.1 does not appear to have explanation. Maybe @ocefpaf can explain why that version is preferred?

It appears that they added a new symbol in 2.8.1 and `master` has some new one as well. Unless there is something else wrong with 2.8 that I don't understand maybe we can build against 2.8 and then set run to require 2.8*.

ref: https://abi-laboratory.pro/tracker/timeline/freetype/

PR #64, opened 2018-04-19
Try building osx and win

ocefpaf:
@isuruf another test for the CircleCI OS X feature.


msarahan:
Yikes, that is an exceedingly confusing bit of text.  My guess based on the output would be that it's using 10.13 over 10.9, but we'd definitely need to see the output to see what actually came out.

mingwandroid:
> I'm not sure which sdk is being used here. Maybe we should wait until we move to the anaconda compilers.

We've upgraded to Qt 5.9.5 now, there are a few [rough edges](https://github.com/ContinuumIO/anaconda-issues/issues/9199) that I am filing down but nothing too major. It would be good to wait for that I think.

jakirkham:
Could try setting `QMAKE_MACOSX_DEPLOYMENT_TARGET` to `10.9`. Doesn't look like we are doing that already (based on a quick skim).

jakirkham:
In case this of interest and/or you have ideas, please feel free to chime in, @stuarteberg.

ccordoba12:
Given that Appveyor didn't even start to build WebEngine, I think you should ask for 3 hours there, perhaps four.

> It's using the MacOSX10.13.sdk but will attempt to make something that works on 10.9. This is unlikely to end well.

Agreed with Ray here.

> Could try setting QMAKE_MACOSX_DEPLOYMENT_TARGET to 10.9. Doesn't look like we are doing that already (based on a quick skim).

I think the minimum macOS version (and SDK, of course) needed by Qt 5.9 is 10.11. Ray can correct me if I'm wrong.

mingwandroid:
You need to use the 10.11 SDK (because templates were added to some Cocoa containers) but you can 
 (and should) set `-mmacosx-version-min=10.10`.

Interestingly, you *can* build all of this on macOS 10.9 provided you use our compilers and put the macOS SDKs in place (and locate them via `CONDA_BUILD_SYSROOT`).

mingwandroid:
Really guys, don't waste your time with Qt 5.6, there's just no point. Defaults has 5.9 and I'm more than happy to work on getting all of my changes merged with conda-forge, but you need to use my compilers.

ocefpaf:
> Really guys, don't waste your time with Qt 5.6,

Don't worry! The effort is not to build `Qt 5.6` but only to test CircleCI as the OS X provider :smiley: 

> Defaults has 5.9 and I'm more than happy to work on getting all of my changes merged with conda-forge

I'll try to port some of your changes here to test but I'm not planning to merge this.
So I'll probably take you on your offer :wink: 

> , but you need to use my compilers.

That is already decided BTW. We are moving to defaults compilers.

mingwandroid:
> We'll need to update the dependencies to use the newer compilers before using them here right?

Well, there are practically no hardcoded dependencies anymore (thanks to using conda_build_config.yaml) and those that are present conda-forge already have. For the cdt packages and the compilers you'd need the defaults channel to appear (last I guess):

https://github.com/AnacondaRecipes/qt-feedstock/blob/master/recipe/meta.yaml

.. so you should be able to test the build. You will need to set `CONDA_BUILD_SYSROOT` to where-ever your macOS SDK is (we do this in conda_build_config.yaml; try to use a 10.11 one if you can).

PR #72, opened 2018-08-02
Windows build

isuruf:
Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`


gillins:
> I'll ping them

That would be great. I haven't had any luck with `vc9` - VS 2008 Express is 32 bit only. Seems conda-build is really only trying to support AppVeyor (not local builds) now that they have dropped support for VS For Python 2.7 (which was 32 and 64 bit).

ocefpaf:
> Seems conda-build is really only trying to support AppVeyor

I don't think that is true b/c their main use is non-AppVeyor (`conda-forge` on the other hand is virtually 100% focused on AppVeyor). But things did change quite a bit and maybe @msarahan can shed some light.

PS: I'll keep you in the look if AppVeyor say something about extending our time here.

isuruf:
Btw, appveyor increased time limit from 60 to 90 minutes for conda-forge a month before and I don't think they increase the limit more than 90 minutes

isuruf:
@jakirkham, isn't the timeout specified for an appveyor organization?

jakirkham:
I don't know. Easy to test though.

PR #75, opened 2018-08-29
[just testing] Qt5.11

hmaarrfk:
Qt 5.11 claims to have better high DPI support. I'm curious to see if it compiles.

I can compile it with the provided options on ubuntu 16.04, but not on my conda-builld. maybe there something that the CI's setup that I don't.

I also wonder what patches are still needed for OSX and Windows. None of the 5.6 patches were applicable for 5.11, at least on Linux.

Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)


mingwandroid:
Doing things like removing C++11 seem a bit arbitrary to me. Any explanation as to why?

ccordoba12:
> Also why add another dependency to Qt? It is not needed

I agree with Ray here. There's no need to make things more complex than they already are.

hmaarrfk:
@mingwandroid I'll definitely look at your recipe. Looks really elaborate. My OSX build seems to have trouble with some posix stuff. Maybe it was also related to your version bump to [OSX 10.11](https://github.com/AnacondaRecipes/qt-feedstock/blob/master-5.11/recipe/build.sh#L190)

I added that dependency because OSX failed once on not being able to find `re2/re2.h`  <-  I guess this is what PCRE is for.

c++11 flag was giving me an error. I guess it was removed. Ray's recipe seems to have removed it too.

Do you generally keep the historical patches?

Do we need to rebuild Qt on minor library version bumps? I guess so as they might add or remove symbols from their shared libraries.

mingwandroid:
> Do we need to rebuild Qt on minor library version bumps?

It depends .. we only support LTS releases and they only come every so often but they do end up containing more changes than I'd consider appropriate for LTS, personally. Having said that, the updates frequently contain important security updates so not updating isn't really an option. Of course you can read the detailed release notes at each point release to make a judgement on whether it's important or not. For sure we skip some point releases as there's not time to catch them all still the goal is to provide the latest possible (with *various* caveats!).

mingwandroid:
> Do you generally keep the historical patches

When we force push we always try to create a backup. Different people have different conventions, mine is to push to e.g. master-5.9.6 before a force push .. but I don't see much value in old patches personally, the newest ones will be the ones to use, I'd hope.

In the case of most packages adding pcre as a dep is the right way to go, but Qt is tricky due to its size, the number of deps it already has and its complexity.

hmaarrfk:
@mingwandroid where do you define the variable qtwebkit that you use for selectors?

mingwandroid:
In the top level conda_build_config.yaml in AnacondaRecipes/aggregate

hmaarrfk:
@conda-forge-linter please be quiet

hmaarrfk:
hmm, now missing qmake :\ weird. it seems to work on my laptop (at least until missing the xcb_shm symbols)

leycec:
**Qt 5.12 (LTS) is due to [ship in November](https://blog.qt.io/blog/2018/02/22/qt-roadmap-2018/).**

The first stable release of Qt for Python (i.e., PySide2) [only officially supports Qt >= 5.11](https://blog.qt.io/blog/2018/06/13/qt-python-5-11-released/). For that reason alone, bumping this feedstock to support Qt 5.12 on release is a critical milestone with significant downstream implications. *Marketing speak translation:* a Qt >= 5.11 conda-forge package enables us to bump the [geriatric Qt 5.6-specific PySide2 conda-forge package](https://github.com/conda-forge/pyside2-feedstock), which is [badly in need of attention](https://github.com/conda-forge/pyside2-feedstock/issues/21). This isn't your average Qt LTS bump. <sup>*actually, it probably is*</sup>

PySide2 is a mandatory dependency of [our multiphysics biology simulator](https://github.com/conda-forge/betsee-feedstock). Thanks to that, I might be able to connive management (*hi, guys!*) into granting me sufficient time to devote to this pull request. If so, what would be the best expenditure of my time here? Or would it possibly be more prudent to bide my time in perfidious silence until Qt 5.12 officially drops?

...'cause I can do that, too.

**tl;dr:** Let me know if you want me to sacrificially do free work for you guys on this pull request.

hmaarrfk:
@leycec Yeah, I've only been able to compile this locally on my own computer. I feel like I should probably just close this PR and let you open your own. I mostly didn't want to setup the CI's on my own fork.

leycec:
@hmaarrfk Sound advice, for you are wise. You've invested a non-trivial amount of effort here; I'll *definitely* base my work on this changeset. You're probably almost there already, in fact. Which is great, for I am lazy.

If it's not too much to ask, would you mind keeping this open? It'll probably be a week or five <sup>*damn you, Murphy's Law*</sup> until I have a passing solution. In the meanwhile, an open pull request on this topic could drive some useful discussion.

Thanks for all the excellent volunteerism. My crystal ball shows positive karma in your future. :crystal_ball:

hmaarrfk:
Yeah, this will stay open. I think what was a source of problems were the older compilers. Since conda-forge was going to upgrade compilers, I figured I would wait until the upgrade passed. I think qt just got rebuilt anyway, so it should be ready for an upgrade.

mingwandroid:
We don't know which way to go with qt-feedsock yet, but I'd like to get our recipes aligned.

leycec:
@hmaarrfk and @mingwandroid: You've both done an *astonishing* amount of work here – especially @mingwandroid, whose [recipe appears to be slightly more authoritative](https://github.com/AnacondaRecipes/qt-feedstock/blob/master-5.11/recipe/meta.yaml). <sup>*maybe?*</sup> I'm flabbergasted, actually. We absolutely must preserve as much of this as possible. Are either of you Qt contributors or developers, by any chance?

On perusing [@hmaarrfk's changelog](https://github.com/conda-forge/qt-feedstock/pull/75/files), I'm left with a few inevitable questions:

1. **The number of patches has significantly increased.** The [current recipe](https://github.com/leycec/qt-feedstock/blob/master/recipe/meta.yaml) requires 10 patches, which is already a bit much; [@mingwandroid's recipe](https://github.com/AnacondaRecipes/qt-feedstock/blob/master-5.11/recipe/meta.yaml) requires **16 patches.** That isn't necessarily a good thing. Each additional patch increases the conda-forge maintenance burden on version bumps. Moreover, more than a few of these patches appear to be a bit... *dodgy.* As example, the header for [`0018-Revert-macOS-ensure-mouse-events-from-the-global-mon.patch`](https://github.com/conda-forge/qt-feedstock/pull/75/files#diff-f5fc2b85cc0940d818d056278c933ca3R9) explicitly cautions that: `May break apps that use pop-up windows.` Are all or even most of these patches required and desired?
1. **Which feedstock branch should my branch be rebased from:** [@hmaarrfk's `qt5.11` branch](https://github.com/hmaarrfk/qt-feedstock/tree/qt5.11) or [@mingwandroid's `master-5.11` branch](https://github.com/AnacondaRecipes/qt-feedstock/tree/master-5.11)? My working assumption is the latter, but I'd hate to summarily discard @hmaarrfk's earnest efforts.
1. **What specifically is broken in your branch, @mingwandroid?** I note, for example, that your most recent commit reads ["Fix build on ubuntu, add cups (WIP, broken)."](https://github.com/AnacondaRecipes/qt-feedstock/commit/bd123a8b889adc3608b53e796e8c4d3434a1faed) *Oh, boy.*
1. **Should I even be doing this?** I have sufficient free time to burn... for the moment. I'm concerned, however, that this may not necessarily be a sane use of that time – especially if @mingwandroid already has a prospective pull request in the works.
1. If I *should* be doing this, **is there some way that we can distribute the workload over the three of us?** What appears to be an insurmountable task for a single developer should be tractable for the three of us, assuming we can cleanly work in parallel on this without continually stepping on each other's toes. And I have pretty big toes.
1. Pertinent to that last question, **would it perhaps make more sense for the three of us to submit pull requests against [@mingwandroid's `master-5.11` branch](https://github.com/AnacondaRecipes/qt-feedstock/tree/master-5.11) rather than here?** Since @mingwandroid already has a branch that purports to behave as expected on at least one platform (i.e., macOS), it might make more sense to unify our efforts there first.

In short, I'm rapidly over my head. Having architected the [PySide2 and shiboken2 ebuilds for Gentoo Linux](https://github.com/gentoo/qt/tree/master/dev-python/pyside), I'm intimately familiar with the Qt build system. That said, there's a significant difference between packaging a single Qt project for a single platform versus the entirety of Qt for all supported Anaconda platforms.

Help, for I am drowning in a sea of ponderous information! :ocean:

hmaarrfk:
@mingwandroid is definitely the authority. Mine was kinda this one thing that worked on my own computer. 

As for I this is worth it, it might be better to simply catch up with defaults.

I think Qt will have an other LTS release soon. Maybe we should way for that after catching up?

Ultimately, I wouldn’t waste any time with the old compilers. I’m not sure how to Mark packages to use only the new compilers. The migration seems to be on track now that qt got through. Hopefully the rest of the python ecosystem will fall through soon.

mingwandroid:
I have contributed fairly minor patches to Qt and QtCreator in the past but wouldn't consider myself a Qt developer as such.

Most of the macOS patches are due to either:

1. Qt upstream making decicisions about the minimum supported OS version that conflicts with our minimum supported OS version (these do have a high risk of breaking something).
2. Qt refusing to build with anything but system compilers on macOS. We cannot use these because we need to use our own compilers for openmp support. When trying to mix binaries generated by system compilers with our own we risk incompatibility (exception handling, C++ structure mismatches) so instead we build *everything* with our own.

For any patch in particular, demonstrate a bug and I'll do my best to fix it. I feared there may be an issue with that patch but no one has demonstrated that so far so I may have been wrong. I'm always happy to be able to remove patches but upstreaming to Qt is difficult. The will not be interested in patches to regain backwards compatibility for example and (rightly so) are strict about quality. Some of my old Qt patches took weeks to pass review. This is something I don't have too much time for at present.

My WIP branches will tend to be in broken states. I hack on them when I'm waiting for other things to finish usually, then move back as soon as I can.

From an AD perspective, moving beyond Qt 5.9 requires (AFAIR) macOS 10.11 or later and my attempts to undo that restriction failed (there were too many changes required, mostly to QtWebEngine), so until we declare 10.11 the minimum supported macOS I'm unlikely to spend a lot of time on Qt > 5.9. Now that *could* happen sometime soon-ish anyway.

For the cups thing, It'll be minor, and in fact might work if you add the cups CDT package may I since added 

We want to sync our recipes as much as possible, CF have decided to use their own X11 packages though while AD uses system X11 packages. We need to hide that difference behind a mutex package and a metapackage in the future I think. Unfortunately I will probably have to do the work to re-sync our recipes and I'm always pretty busy.

But please join the effort(s) here! (I'd recommend tackling the CDT vs X11 difference between AD and CF for the 5.9.x recipe as your first task). Qt is amongst the trickiest packages we build so having people with experience of it join in is always welcome.

I wouldn't get too hung up on the patch count. Clealry I could've combined all my changes into 1 patch and that'd look good numerically, but it'd be a terrible thing to do. Most of the patches are simple and obvious, but unfortunately often un-upstreamable here.

leycec:
## hmaarrfk for Great Justice

> @mingwandroid is definitely the authority. Mine was kinda this one thing that worked on my own computer.

I suspected that was the case. Still, it's likely that you made significant advancements towards a working Qt 5.11 package as well, @hmaarrfk. Do you happen to recall what those might have been? Moreover, would you have the time and/or inclination to submit a pull request against [@mingwandroid's master-5.11 branch](https://github.com/AnacondaRecipes/qt-feedstock/tree/master-5.11) integrating your changes with his?

> As for I this is worth it, it might be better to simply catch up with defaults.

I... don't know what this sentence means. Admittedly, [I'm not the sharpest pencil in the drawer](https://dan.hersam.com/lists/not_bright.html).

What does "defaults" refer to, in this case? `</sigh>`

> Ultimately, I wouldn’t waste any time with the old compilers. I’m not sure how to Mark packages to use only the new compilers. The migration seems to be on track now that qt got through. Hopefully the rest of the python ecosystem will fall through soon.

I... don't know what this paragraph means, either. Yet again, I get the slimy feeling that I'm the wrong volunteer for this thankless job. Let's assume for the sake of naïve curiosity that I actually know what I'm talking about, which I don't. Please define the above undefined terms.

Specifically, what do the phrases **old compilers,** **new compilers,** and **migration** mean in the context of this discussion? Who is migrating what from what to what, exactly? I maintain a few conda-forge feedstocks, but am clearly out-of-the-loop with respect to low-level conda-forge developments. `</elongated_sigh>`

Life is simple in Gentoo Linux Land. Life is not quite so simple in Anaconda Land. I am saddened.

> I think Qt will have an other LTS release soon. Maybe we should way for that after catching up?

**Right.** Qt 5.12 (LTS) is scheduled to release on [November 29th](https://wiki.qt.io/Qt_5.12_Release), less than two weeks away. A Qt 5.12 Beta was released [more than a month ago](https://blog.qt.io/blog/2018/10/05/qt-5-12-lts-beta-released), which gives us a reasonably stable target. That said, retargeting @mingwandroid's hard work from Qt 5.11 to 5.12 could prove... non-trivial. In particular, a number of those patches seem unlikely to apply cleanly to Qt 5.12. <sup>Of course, I'm usually wrong about everything.</sup>

For everyone's collective sanity, preserving @mingwandroid's contributions would seem to be of higher priority than refactoring this changeset to support 5.12. That can come later, right? First, we should just get this working. The perfect is the enemy of the good. <sup>...did I mention that I'm usually wrong about everything?</sup>

## mingwandroid for the Win

@mingwandroid, you are an encyclopedia of wisdom in human guise. :book:<sup>←*this is you*</sup>

> I have contributed fairly minor patches to Qt and QtCreator in the past...

**Huzzah!** I *knew* one or both of you had to be hip-deep into Qt Company (QtC) development. That mind-bending patchset was the dead giveaway.

> ...but wouldn't consider myself a Qt developer as such.

Humility will get you nowhere. By compare to the rest of this motley crew, you're the *de facto* Qt developer in residence. Be proud, for you are singular.

> Qt upstream making decicisions about the minimum supported OS version that conflicts with our minimum supported OS version (these do have a high risk of breaking something).

**Right.** The Qt Company (QtC) often makes laughably nonsensical decisions. (See: *"What is renaming PySide2 to "Qt for Python," Bob?"*) I fully support pushing back against bad upstream choices – assuming we can, of course. Which we sometimes can.

> Qt refusing to build with anything but system compilers on macOS. We cannot use these because we need to use our own compilers for openmp support. When trying to mix binaries generated by system compilers with our own we risk incompatibility (exception handling, C++ structure mismatches) so instead we build everything with our own.

**Right.** OpenMP support absolutely takes precedence in the analytics space that Anaconda occupies. Again, I fully support pushing back against bad upstream choices. :+1: 

> For any patch in particular, demonstrate a bug and I'll do my best to fix it.

No, you're quite right. You've clearly thought through the implications of this patchset and are clearly in the right. Happily, I have no bugs to demonstrate. Allow me to apologize for derailing the discussion. Cue sad face. :disappointed: 

> They will not be interested in patches to regain backwards compatibility...

Well, that further confirms my suspicions of the Qt Company. Backwards compatibility should *always* be their first priority, especially if somebody else (e.g., you) is willing to volunteer working patches. *Urgh!*

> Some of my old Qt patches took weeks to pass review. This is something I don't have too much time for at present.

**Absolutely.** Myself, I have a strict policy of *never* volunteering for for-profit corporations. If they can't be bothered to remunerate my valuable time, I can't be bothered to generate profit for them. <sup>*that's just how we roll*</sup>

> My WIP branches will tend to be in broken states.

Your WIP branches are the "best thing going," at the moment. If you wouldn't mind hacking on them a bit further, that'd be fantastic. Your endless volunteerism is the linchpin here. I just wish we had some means of repaying your continued generosity.

Emoji will have to do. Here's a delicious Thanksgiving turkey for you. :turkey: 

> From an AD perspective...

**Uh oh.** Here's where I completely fall off the collective bandwagon. If you wouldn't mind, please define the following acronyms for this crass simpleton:

* **AD.** After Death? After Dark? Anaconda Desktop? Anaconda Cloud? Two of these are likelier than the others.
* **CDT.** Central Daylight Time? Common Driver Trainer? Common Desktop Toolkit? Eclipse's C[++] Development Toolkit? I am clueless.
* **CF.** Conda-forge! This one has *got* to be conda-forge, right? <sup>*shoot me now*</sup>

> From an AD perspective, moving beyond Qt 5.9 requires (AFAIR) macOS 10.11 or later and my attempts to undo that restriction failed (there were too many changes required, mostly to QtWebEngine),

This is me shaking my arthritic fist at Apple, Google, and the Qt Company. As much as I loathe Apple's [forced upgrade policy promoting planned obsolescence and conspicuous consumption](https://phys.org/news/2017-12-iphones-apple-owners.html), El Capitan (10.11) *does* appear to be a reasonable requirement in late 2018. Surprisingly, El Capitan officially supports [ancient iMac and MacBook Pro models from mid-2007](https://www.macworld.com/article/2988074/os-x/apple-launches-os-x-10-11-el-capitan-heres-how-to-get-it.html) – more than a decade old.

It seems doubtful that hardware *that* geriatric would perform well against a modern development workload, anyway. Especially under macOS. Gentoo or Arch Linux? Now that's another story, of course. But given that this *is* the most infamously expensive desktop and mobile platform of all time, well... Apple consumers really have no expectation of platform longevity or forward compatibility.

**tl;dr:** *El Capitan (10.11) seems reasonable. Please seek a second opinion. I am not to be trusted.*

> CF have decided to use their own X11 packages though while AD uses system X11 packages. We need to hide that difference behind a mutex package and a metapackage in the future I think... I'd recommend tackling the CDT vs X11 difference between AD and CF for the 5.9.x recipe as your first task.

_Yay!_ A meaningful first milestone. Sadly, I have no idea what AD, CF, or CDT mean. I know what X11 means, of course, but... that's unlikely to get us anywhere.

I'm also unclear as to what distinguishes a mutex package from a metapackage. By definition, aren't all mutex packages also metapackages? To quote the [only authoritative discussion I could find on the subject](https://github.com/conda-forge/conda-forge.github.io/issues/587#issuecomment-390057588):

> Let me sketch it out. The key is the creation of a set of mutually exclusive "mutex" metapackages: the same package name, different build strings. There is one per variant. The packages have no dependencies. Each of the variants that you build includes its appropriate metapackage as a dependency.
>
> To enable a strong "default" preference, the *non-default* packages all include a `track_package` value. This ensures that conda's "feature minimization" protocol will prefer the default package. It's not good enough to just give the default metapackage a lower build or version number, because that will generally not be given sufficiently high priority. The feature minimization, on the other hand, has a very high priority.
>
> Now, in order to handle the *manual selection* of a non-default package, you can create a set of *selector metapackages.* Each of these packages has a *different name,* and include the corresponding mutex metapackage as a dependency. That way, if you install a selector metapackage, constraint conflicts will prevent it from being replaced; and this forces that variant to be pinned.
>
> Now, selector metapackages may not be necessary in this case, *if* there are a natural set of packages that can serve as selectors. For example, perhaps the different MPI implementations can serve as selectors themselves. If so, just make sure they have a dependency on their corresponding mutex package.

Does any of the above apply to the discussion at hand? Most of my difficulties surrounding this discussion appear to revolve around the fact that *none of this appears to be documented anywhere.* Prominent search engines fail to return hits for "conda feature minimization protocol," for example – leaving me largely clueless as to what, exactly, I should be doing and why.

This is where I mournfully sigh. I am ignorant and willing to help, but only one of those two things is useful.

mingwandroid:
@leycec, you *always* cheer me up (maybe I'm always grumpy?) with your comments, thanks!

Some definitions ..

AD = Anaconda Distribution (now largely based on conda-forge and increasingly compatible).
CF = conda-forge.
CDT = Core Dependency Tree (see more at https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html or https://conda.io/docs/user-guide/tasks/build-packages/compiler-tools.html#cdt-packages)

I believe I was incorrect when I said 10.11 was the min for Qt > 5.9.x, it's actually 10.12, still Apple users seem to update more regularly than other OS users (probably helped by those nagging daily notifications).

I wouldn't thank me too much for my work here as I am an Anaconda, Inc. employee paid to work on this stuff :-) Contributing to conda-forge is now (since we're based on their recipes and compilers etc) much more pain free than it used to be for us Ananconda, Inc. employees. It's also a stated goal for AD to be as compatible as it can be with CF (with some minor other concerns that conflict with that goal; the CDT vs CF X11 issue for example - I've yet to see what'll happen from attempting to swap out AD Qt fot CF Qt; it *might* still work since Qt do a good job of defining their public C++ interface to ensure binary compatibility).

mingwandroid:
On the patch count matter here, it may be good to remember how vast Qt is. [The source archive itself is > 460MB compressed with LZMA!](https://download.qt.io/archive/qt/5.11/5.11.2/single/)

So I think a ratio of LOC vs LO-patch would actually put Qt as one of the least heavily patched packages.

mingwandroid:
> I'm also unclear as to what distinguishes a mutex package from a metapackage

>> The packages have no dependencies

So metapackages will always have dependencies, and mutex packages will never have them.

hmaarrfk:
Qt 5.11 compiled on Ubuntu 16.04 (I think) using some compiler I had installed at the time. It might have been gcc 6. 

Conda forge decided to simultaneously update the compiler they used at the same time as python 3.7 and some r upgrade. This means alot of ci resources were used and are still being used for what essentially amounts to a while district rollout based on free CIs. (I think some donations went toward paying for appveyor) https://conda-forge.org/status/

https://youtu.be/Kamld5Z-xx0

This is a great talk that I think showcases @mingwandroid s work (I don’t know if Michael references the same person) but explains how anaconda and conda-forge decided to compile software. 

Conda-forge’s recent upgrade i think now puts them in line with anacondas compilers


As for the cost of apple laptops, having purchases similarly priced windows machines, I think apple laptops come with many intangible upgrades. These come in the form of great screens, great trackpads, and the fastest ssd.

mingwandroid:
> Well, that further confirms my worst QtC suspicions. Backwards compatibility should always be their first priority, especially if somebody else (e.g., you) is willing to volunteer working patches. Urgh!

I think I understand why they do these things:
1. New h/w features (you cannot support that awful touchbar without using a recent SDK; probably for the best not to support this h/w misfeature; I want function keys, real ones, that you can press and that don't get pressed by mistake when your hand swipes past some touch sensor, ones where you know where the boundary between keys is because the keys are physical).
2. They hitched their compatibility to Google's Blink for QtWebEngine which was a terrible decision IMHO (I would've preferred them to update to webkit 2 instead here).

I just wish they would make macOS SDK level selectable via a configuration knob, but given that QtWebEngine forces recent SDKs I guess from QtC's perspective there's little point in the rest of the Qt libraries supporting an earlier version.

mingwandroid:
> As for the cost of apple laptops, having purchases similarly priced windows machines, I think apple laptops come with many intangible upgrades. These come in the form of great screens, great trackpads, and the fastest ssd.

Apple h/w is only 'ok' now, I wouldn't recommend the new ones to anyone due to the keyboard and touchbar though. My old MBP went back to Apple 4 times in 2 years for extended periods with broken keyboards and an SSD that would randomly disappear from the system. I think the [Windows laptop world has caught up largely](https://www.androidauthority.com/huawei-matebook-x-pro-black-friday-924718/), ok, in this case by taking a load of inspiration from Apple (but ditching the terrible bits), but hey, as Steve Jobs claimed Picasso once said, "Good artists copy. Great artists steal".

mingwandroid:
Thanks for your nice comments but really credit goes to everyone. Mike was talking largely about conda-build features in that YouTube video and he implemented most of that stuff and also did a lot of work on the compilers (we tag teamed that one!)

Nehal, Jonathan and @isuruf have also contributed a lot as has the wider conda-forge community. It's a pleasure to work with everyone here.

hmaarrfk:
@mingwandroid didn't mean to leave anybody out :/ Thanks for sharing more about the project contributors.

@leyec I think I hit really hard roadblocks because of the versions that conda(-forge) tries to target.
My computer has so many libraries installed that it might have been linking to really new ones that qt supports. My final commits states: "start from @mingwandroid recipe" so I likely thought that it was the best starting point :/

hmaarrfk:
@leycec here is how to skip building with the old compilers:
@ SylvainCorlay  said

> @hmaarrfk if it is something that depended on toolchain on old compilers
```
 {% if 'toolchain' in compiler('cxx') %}skip: True{% endif %}  # Only build with modern compilers.
```

ocefpaf:
Let's go with #110. (Unless you really need 5.11 @hmaarrfk, then please re-open this one.)

hmaarrfk:
Totally fine.

PR #79, opened 2018-10-23
Iterate the regro rebuild

pkgw:
This version adds a patch that should get the Qt build system to honor build environment variables more reliably.


pkgw:
Attempts to supersede #78.

pkgw:
Some of the first Linux/GCC7 errors (buried deeeeep in the logs):

```
In file included from dom/Node.h:39:0,
                 from dom/ContainerNode.h:28,
                 from dom/Document.h:33,
                 from inspector/DOMEditor.cpp:38,
                 from inspector/InspectorAllInOne.cpp:30:
../WTF/wtf/ListHashSet.h:269:63: error: return type specified for deduction guide
         template<typename T> static unsigned hash(const T& key) { return HashArg::hash(key->m_value); }
                                                               ^
../WTF/wtf/ListHashSet.h:269:37: error: decl-specifier in declaration of deduction guide
         template<typename T> static unsigned hash(const T& key) { return HashArg::hash(key->m_value); }

In file included from ../WTF/wtf/HashMap.h:24:0,
                 from inspector/InspectorValues.h:35,
                 from generated/InspectorTypeBuilder.h:12,
                 from generated/InspectorFrontend.h:9,
                 from inspector/ConsoleMessage.h:36,
                 from inspector/ConsoleMessage.cpp:35,
                 from inspector/InspectorAllInOne.cpp:28:
../WTF/wtf/HashTable.h: In instantiation of 'WTF::HashTable<Key, Value, Extractor, HashFunctions, Traits, KeyTraits>::AddResult WTF::HashTable<Key, Value, Extractor, HashFunctions, Trait>
../WTF/wtf/ListHashSet.h:720:38:   required from 'WTF::ListHashSet<Value, inlineCapacity, HashFunctions>::AddResult WTF::ListHashSet<Value, inlineCapacity, HashFunctions>::add(const Valu>
dom/Node.h:799:21:   required from here
../WTF/wtf/HashTable.h:828:42: error: 'hash' is not a member of 'WTF::ListHashSetTranslator<WebCore::KURLHash>'

```

pkgw:
The errors from Linux/toolchain_c, also buried deep in the logs:

```
In file included from ../3rdparty/javascriptcore/JavaScriptCore/runtime/RegExp.h:30:0,
                 from ../3rdparty/javascriptcore/JavaScriptCore/bytecode/CodeBlock.h:40,
                 from ../3rdparty/javascriptcore/JavaScriptCore/API/JSCallbackFunction.cpp:32:
../3rdparty/javascriptcore/JavaScriptCore/yarr/RegexJIT.h: In destructor ‘QTJSC::Yarr::RegexCodeBlock::~RegexCodeBlock()’:
../3rdparty/javascriptcore/JavaScriptCore/yarr/RegexJIT.h:65:36: error: ‘jsRegExpFree’ was not declared in this scope
             jsRegExpFree(m_fallback);
                                    ^
In file included from ../3rdparty/javascriptcore/JavaScriptCore/runtime/RegExp.h:30:0,
                 from ../3rdparty/javascriptcore/JavaScriptCore/bytecode/CodeBlock.h:40,
                 from ../3rdparty/javascriptcore/JavaScriptCore/API/JSCallbackFunction.cpp:32:
../3rdparty/javascriptcore/JavaScriptCore/yarr/RegexJIT.h: In function ‘int QTJSC::Yarr::executeRegex(QTJSC::Yarr::RegexCodeBlock&, const UChar*, unsigned int, unsigned int, int*, int)’:
../3rdparty/javascriptcore/JavaScriptCore/yarr/RegexJIT.h:89:88: error: ‘jsRegExpExecute’ was not declared in this scope
         return (jsRegExpExecute(fallback, input, length, start, output, outputArraySize) < 0) ? -1 : output[0];
```

pkgw:
Error from the test stage of macOS/clang below. The build succeeded. I think this just means that we need to add clang as a test requirement:

```
The following NEW packages will be INSTALLED:

    icu:       58.2-h0a44026_1000   conda-forge/label/gcc7
    jpeg:      9c-h1de35cc_1001     conda-forge/label/gcc7
    libcxx:    4.0.1-h579ed51_0     conda-forge/label/gcc7
    libcxxabi: 4.0.1-hebd6815_0     conda-forge/label/gcc7
    libpng:    1.6.35-ha441bb4_1002 conda-forge/label/gcc7
    ncurses:   6.1-h0a44026_1001    conda-forge/label/gcc7
    qt:        5.6.2-he68f1f2_1009  local                 
    readline:  7.0-hcfe32e1_1001    conda-forge/label/gcc7
    sqlite:    3.25.2-h1765d9f_1000 conda-forge/label/gcc7
    zlib:      1.2.11-h1de35cc_1003 conda-forge/label/gcc7

Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
Info: creating stash file /Users/distiller/miniconda3/conda-bld/qt_1540348147662/test_tmp/test/.qmake.stash
x86_64-apple-darwin13.4.0-clang++ -c -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -std=c>
make: x86_64-apple-darwin13.4.0-clang++: No such file or directory
make: *** [main.o] Error 1
```

pkgw:
macOS/toolchain_c has a configuration error; this should be fixable in some straightforward way since this build used to work ...

```
System architecture: 'x86_64'
Host architecture: 'x86_64'
checking for C++11... 
/Applications/Xcode-9.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -c -mmacosx-version-min=10.9 -stdlib=libc++ -isysroot /Applications/Xcode-9.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -m64 -fPIC -I/Users/distiller/miniconda3/conda-bld/qt_1540347631528/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/include -mmacosx-version-min=10.9 -isysroot /Applications/Xcode-9.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -m64 -fPIC -I/Users/distiller/miniconda3/conda-bld/qt_1540347631528/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/include -pipe -stdlib=libc++ -O2 -isysroot /Applications/Xcode-9.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -mmacosx-version-min=10.9 -Wall -W -fPIC  -I. -I../../../mkspecs/macx-clang -o c++11.o c++11.cpp
c++11.cpp:37:4: error: "__cplusplus must be >= 201103L, or one of __GXX_EXPERIMENTAL_CXX0X__ or __INTEL_CXX11_MODE__ must be defined"
#  error "__cplusplus must be >= 201103L, or one of __GXX_EXPERIMENTAL_CXX0X__ or __INTEL_CXX11_MODE__ must be defined"
   ^
1 error generated.
make: *** [c++11.o] Error 1
C++11 disabled.
c++11 support cannot be enabled due to functionality tests!
```

(This build log is actually quite short.)

pkgw:
@jakirkham Yes, that's what it looks like. Just submitted a new version; we'll see if it fixes macOS. Linux looks like it's going to take more digging ... although toolchain_c worked in the past, so hopefully it will also be straightforward to fix with the proper insight.

pkgw:
Currently locally testing whether [spack#4270](https://github.com/spack/spack/pull/4270) can fix Linux/toolchain_c ... potentially leaving Linux/gcc7 as the last hurdle.

pkgw:
And my guess is that the Linux/gcc7 failure is due to CXXFLAGS containing `-std=c++17` by default (as per recent conversation on Gitter). Will try to commandeer another machine and see if that helps ...

mingwandroid:
We do [this](https://github.com/AnacondaRecipes/qt-feedstock/blob/master/recipe/build.sh#L7-L11) in `AD`

pkgw:
@mingwandroid This raises a topic that I don't feel that I understand well — at this point, should the first order of business be to try adopting the AD build scripts when doing these rebuilds? I don't have a good sense for how similar the AD and CF build environments are.

mingwandroid:
By and large, once conda-forge have moved to the new compilers (and even to an extent even during the transition due to careful work by @conda-forge/core) our recipes and scripts should be highly synchronizable, should conda-forge want to, on a case-by-case basis.

For us, the intention is for the community to take the lead wrt recipes and for us to try to help out where we can be useful, so we'll tend to rebase on top of cf for most things and only carry additional patches when necessary.

Of course there'll be points of difference, X11 vs CDT being the obvious one, but for that I hope some metapackages can be used to hide those differences.

pkgw:
I am hopeful that this latest iteration will succeed on all four platforms here. **However**, I did notice one thing: on at least Linux/gcc7, the linking report indicates a dependency on "system libraries" like `/usr/lib64/libfreetype.*`, fontconfig, and so on.

These are also installed as in-prefix dependencies, but I wonder if the linker flags being used are somehow causing the build to favor the versions in `/usr/lib64`. Does anyone know how this works? Will the right dependency emerge in the final packaged executables anyway?

One way to avoid this would be to depend on the C-F X11 packages, eliminating the need to link with *any* library in `/usr`, but the previous version(s) of this package didn't do so.

mingwandroid:
> /usr/lib64/libfreetype.*,

Which Qt .so linked to this? I tried to check the log but a new build is in-progress. My initial thought is you may need: https://github.com/AnacondaRecipes/qt-feedstock/blob/master/recipe/0004-qtwebengine-use-CONDA_PREFIX_include-for-system_libjpeg.patch but I don't know that you got that far?

AD builds of Qt do not link to any system libraries (beyond the X11 ones we specify via CDT packages). The new `--error-overlinking` flag catches any instances of this and we fix them.

mingwandroid:
Maybe: https://github.com/AnacondaRecipes/qt-feedstock/blob/master/recipe/build.sh#L79-L84 and: https://github.com/AnacondaRecipes/qt-feedstock/blob/master/recipe/build.sh#L120

pkgw:
@mingwandroid Here's a sampling of the output from the test build I did:

```
dbus-1.13.0-h4e0c4b3_1000:
    libdbus-1.so.3 (lib/libdbus-1.so.3)

fontconfig-2.13.1-h2176d3f_1000:
    libfontconfig.so.1 (lib/libfontconfig.so.1)

freetype-2.9.1-h3cfcefd_1004:
    libfreetype.so.6 (lib/libfreetype.so.6)

[...]

    liblzma.so.5 (lib/liblzma.so.5)

zlib-1.2.11-h14c3975_1003:
    libz.so.1 (lib/libz.so.1)

system:
    libICE.so.6 (/usr/lib64/libICE.so.6)
    libSM.so.6 (/usr/lib64/libSM.so.6)
    libX11-xcb.so.1 (/usr/lib64/libX11-xcb.so.1)
    libX11.so.6 (/usr/lib64/libX11.so.6)
    libXau.so.6 (/usr/lib64/libXau.so.6)
    libXcomposite.so.1 (/usr/lib64/libXcomposite.so.1)
    libXcursor.so.1 (/usr/lib64/libXcursor.so.1)
    libXdamage.so.1 (/usr/lib64/libXdamage.so.1)
    libXext.so.6 (/usr/lib64/libXext.so.6)
    libXfixes.so.3 (/usr/lib64/libXfixes.so.3)
    libXi.so.6 (/usr/lib64/libXi.so.6)
    libXinerama.so.1 (/usr/lib64/libXinerama.so.1)
    libXrandr.so.2 (/usr/lib64/libXrandr.so.2)
    libXrender.so.1 (/usr/lib64/libXrender.so.1)
    libatk-1.0.so.0 (/usr/lib64/libatk-1.0.so.0)
    libc.so.6 (/lib64/libc.so.6)
    libcairo.so.2 (/usr/lib64/libcairo.so.2)
    libdl.so.2 (/lib64/libdl.so.2)
    libexpat.so.1 (/lib64/libexpat.so.1)
    libfontconfig.so.1 (/usr/lib64/libfontconfig.so.1)
    libfreetype.so.6 (/usr/lib64/libfreetype.so.6)
    libgdk-x11-2.0.so.0 (/usr/lib64/libgdk-x11-2.0.so.0)
    libgdk_pixbuf-2.0.so.0 (/usr/lib64/libgdk_pixbuf-2.0.so.0)
    libgio-2.0.so.0 (/lib64/libgio-2.0.so.0)
    libgmodule-2.0.so.0 (/lib64/libgmodule-2.0.so.0)
    libgthread-2.0.so.0 (/lib64/libgthread-2.0.so.0)
    libgtk-x11-2.0.so.0 (/usr/lib64/libgtk-x11-2.0.so.0)
    libm.so.6 (/lib64/libm.so.6)
    libpango-1.0.so.0 (/usr/lib64/libpango-1.0.so.0)
    libpangocairo-1.0.so.0 (/usr/lib64/libpangocairo-1.0.so.0)
    libpangoft2-1.0.so.0 (/usr/lib64/libpangoft2-1.0.so.0)
    libpixman-1.so.0 (/usr/lib64/libpixman-1.so.0)
    libpng12.so.0 (/usr/lib64/libpng12.so.0)
    libpthread.so.0 (/lib64/libpthread.so.0)
    libresolv.so.2 (/lib64/libresolv.so.2)
    librt.so.1 (/lib64/librt.so.1)
    libselinux.so.1 (/lib64/libselinux.so.1)
    libuuid.so.1 (/lib64/libuuid.so.1)
    libxcb.so.1 (/usr/lib64/libxcb.so.1)
    linux-vdso.so.1 ()

not found:


+ exit 0
```

mingwandroid:
@pkgw, can you remove the calls to `conda inspect`. It is deprecated by `check_overlinking()` function which provides richer and more readable information, provided the build gets that far.

pkgw:
The CI failure was a transient one. The Linux/toolchain build has succeeded in the past so I think things are basically good in terms of builds finishing without errors. But there needs to be some checking that the Linux packages have good rpaths and are linking against the right libraries.

pkgw:
Hmm, yes, it looks like there's a problem. If I install the Linux/gcc7 package on my machine, the GUI programs all segfault. A few of the plugins are picking up system libraries:

- platformthemes/libqgtk2.so
- PyQt5/libpyqt5qmlplugin.so
- designer/libpyqt5.so

I think what's happening is that my inclusion of `-L/usr/lib64` and `-Wl,-rpath,/usr/lib64` is causing the builds to pick up extra system libraries that are messing things up. I added those flags before I came up with the current scheme for consistently passing flags to the build system, so maybe I can take them out now ...

mingwandroid:
The linux compilers and linkers have a feature called sysroot poisoning whereby they pretend /usr is empty. I've been mulling over enabling that but haven't so far. For AD, there's never any reason to find anything in there, everything system-y is done via CDT packages in the compiler's sysroot instead.

pkgw:
Hmm, the most recent released conda-forge packages of qt also pick up the sames kind of system libraries, and the `conda inspect` reports from the Linux build ([log](https://circleci.com/gh/conda-forge/qt-feedstock/106)) show the same linkages to system libraries. It's not obvious to me what's different, but the GUI programs aren't segfaulting. Anyone from @conda-forge/qt have experience with these matters?

mingwandroid:
I'm doing a lot of work on Qt 5.9.7 at present, maybe we could wait to see where the chips fall? I'm happy to try to make the CF variant link to your X11 packages.

pkgw:
@mingwandroid Yeah, I might try a few more things, but it would be great if your work on the 5.9 series clarified our thinking about the best strategy here. I am *pretty* sure that if we remove the `yum_requirements.txt` and try to link with CF X11, Linux/GCC7 will be fine. (BTW, I checked and Linux/toolchain still works.) If CF can't make use of the CDT packages (which I think is the case?), I have a hunch that that will be the best path forward.

But again, I haven't been involved in the Qt nitty-gritty before, so maybe other CF veterans are more familiar with how the packaging interacts with system libraries here.

mingwandroid:
There's no technical reason conda-forge cannot use the CDT packages. They're freely available to all as are the recipes and `conda skeleton rpm` can be used to create new ones for a range of systems.

They are never installed at runtime on an end users machine, they are used to define the ABI for any system shared libraries we link against. They also allow us to keep track of which packages need system packages and they open us up later to being able to suggest installing the distro-equivalent package at install time should we determine a system DSO is missing.

So CDTs are used in the X11 and gl stacks, but occasionally a few can be used for other purposes. For example in R I use jdk CDT packages to guarantee binary compatibility. Using that instead of whatever happens to be installed on the system or our own openjdk packages at build time would result in a R interpret (well RJava really) that can only be used with modern jdks and we decided we didn't want to exclude people's legacy Java stacks from interoperating with AD.

Personally I'd prefer if conda-forge standardise on own X11 or system X11 rather than mixing (I don't know the SONAME it symbol versioning strategy of X11 but it doesn't look great from what I have seen) so for this recipe the maintainers need to make a choice.

pkgw:
Decided to go ham and swap over to using the CF X11 packages. Will do some local testing to see if/how these work.

pkgw:
OK, my local builds of the Linux packages both have working GUI programs. Next, let's see what the CI says ...

pkgw:
Builds are green and the Linux packages work in my local tests ... @cona-forge/qt, any feedback on all this?

isuruf:
In the log I see, `-isysroot /Applications/Xcode-9.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk`

pkgw:
@isuruf I don't understand — does that indicate a problem?

pkgw:
@isuruf Any follow-up here? This is one of the biggest blockers in the compiler migration project.

ccordoba12:
I'm +1 to merge, but could we remove the Azure Pipelines stuff so this PR is not flagged as failed?

recipe/build.sh
```diff
@@ -1,13 +1,77 @@
...
+    export LDFLAGS="$LDFLAGS -Wl,-rpath-link,$PREFIX/lib"
+    export CPPFLAGS="$CPPFLAGS -DXK_dead_currency=0xfe6f -DXK_ISO_Level5_Lock=0xfe13"
+    export CPPFLAGS="$CPPFLAGS -DFC_WEIGHT_EXTRABLACK=215 -DFC_WEIGHT_ULTRABLACK=FC_WEIGHT_EXTRABLACK"
+    export CPPFLAGS="$CPPFLAGS -DGLX_GLXEXT_PROTOTYPES"
+else
+    compiler_mkspec=qtbase/mkspecs/common/clang.conf
+    flag_mkspec=qtbase/mkspecs/macx-clang/qmake.conf
+
+    export LDFLAGS="$LDFLAGS -Wl,-rpath,$PREFIX/lib"
+    export CXXFLAGS="$CXXFLAGS -std=c++11"
```
ccordoba12: This is a significant divergence from `defaults`. Why not leave Qt to set its own flags?

recipe/build.sh
```diff
@@ -94,8 +145,6 @@ if [ $(uname) == Darwin ]; then
                 -headerdir $PREFIX/include/qt \
                 -archdatadir $PREFIX \
                 -datadir $PREFIX \
-                -L $PREFIX/lib \
-                -I $PREFIX/include \
```
ccordoba12: Why this removal?

recipe/build.sh
```diff
@@ -25,8 +89,6 @@ if [ $(uname) == Linux ]; then
...
```
ccordoba12: And this removal?

isuruf:
@pkgw, it means that the 10.13 sdk is used and the binary built here might not run on earlier macos versions

looooo:
@isuruf can we control the macos sdk version on circle-ci? If not should we disable `osx_c_compilertoolchain_ccxx_compilertoolchain_cxx`?

recipe/meta.yaml
```diff
@@ -52,6 +53,14 @@ requirements:
     - sqlite
     - libxcb  # [linux]
     - dbus  # [linux]
+    - xorg-libice      # [linux]
+    - xorg-libsm       # [linux]
+    - xorg-libx11      # [linux]
+    - xorg-libxext     # [linux]
+    - xorg-libxrender  # [linux]
+    - xorg-libxt       # [linux]
+    - xorg-inputproto  # [linux]
```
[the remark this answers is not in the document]
pkgw: I didn't want to make this change, but this was the only way I could get it to build everywhere and not crash. I could get it building against the system X libraries on Linux, but the resulting binaries segfaulted in my testing.

recipe/build.sh
pkgw: With the new way that compiler flags are set in this PR, these flags are no longer necessary — they're subsumed into the modifications of the mkspec files.

recipe/build.sh
pkgw: See above. (Marking this as "resolved" now since this and the above are identical points of discussion.)

pkgw:
@isuruf I thought the whole point of `MACOSX_VERSION_MIN` was that you could compile against a new SDK, but still ensure that your binaries would be compatible with older OSs?

@ccordoba12 I am not even sure where the Azure stuff is coming from! I don't see any files that look like magic Azure files. I certainly don't believe there's anything Azure-related in this PR.

recipe/build.sh
pkgw: Well, I guess my overall philosophy is, if you're trying to use a consistent set of compiler flags in your ecosystem and that ecosystem contains a complex, widely-used C++ library, having that library ignore your flags and do its own thing is risky from an ABI standpoint.

But if `defaults` has a working recipe it does seem wise to hew closely to it. I was just building off of what we had here. The choice of system/CDT vs. C-F X.org X11 libraries could make a difference, but I agree that ideally the PR here wouldn't change the status quo. A related factor was that I don't know how to get a C-F recipe to use CDT packages, although I must admit I didn't actually try to figure that out.

ccordoba12:
> I am not even sure where the Azure stuff is coming from!

I know that's not your fault. I think that was activated by @mariusvniekerk.

mariusvniekerk:
Yeah unfortunately azure reports failures if those files don't exist. We've been testing a bit with this feedstock. You can safely disregard those failures

mingwandroid:
>  I thought the whole point of MACOSX_VERSION_MIN was that you could compile against a new SDK, but still ensure that your binaries would be compatible with older OSs?

This has never worked reliably. Besides, Apple have yet to open source .tbd support for ld64 so there's no way to use newer SDKs at present.

recipe/build.sh
mingwandroid: There's no great complexity to using CDT packages, you just include them and they should work, in theory. It's also really easy to create and build your own (someone needed libusb recently and did this). Very occasionally you run into header conflicts between new stuff and old stuff but I've not run into anything too difficult to solve or work-around.

isuruf: @pkgw, `-std=c++17` is already set in the new compiler. Can you do this only for toolchain build? Also you can remove the line above (LDFLAGS) as it's done by both toolchain and anaconda compilers.

recipe/meta.yaml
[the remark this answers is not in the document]
ccordoba12: Ok, no problem.

recipe/build.sh
ccordoba12: Ok, thanks for the explanation.

recipe/build.sh
ccordoba12: @conda-forge/qt, I think a decision needs to be done about this. Should we proceed with the Xorg packages or should we try to use the CDT ones?

My vote is for using the Xorg ones given that a) @pkgw has already make Qt compile with them; and b) I guess very few people has time to start creating the CDT ones.

mingwandroid:
> My vote is for using the Xorg ones given that a) @pkgw has already make Qt compile with them; and b) I guess very few people has time to start creating the CDT ones

Not sure what you mean here. I've already done the work, but I'm also very happy to see how things shake out here using cf x11.

I'll try to make sure we can still share recipes somehow when I can.

ccordoba12:
> I've already done the work

I know, I meant creating recipes for them in conda-forge.

> I'll try to make sure we can still share recipes somehow when I can.

So you agree with merging this one?

mingwandroid:
> So you agree with merging this one?

Absolutely!

ccordoba12:
Ok, merging it is! Thanks a lot @pkgw!!

PR #92, opened 2019-01-29
Repin with Gstreamer 1.14

croth1:
…nda-forge-pinning 2019.01.29

Checklist
* [x] Used a fork of the feedstock to propose changes
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

### DIFF
--- a/.ci_support/linux_.yaml
+++ b/.ci_support/linux_.yaml
@@ -15,7 +15,7 @@ fontconfig:
 freetype:
 - 2.9.1
 gstreamer:
-- '1.12'
+- 1.14.4
 icu:
 - '58'
 jpeg:
@@ -33,8 +33,6 @@ pin_run_as_build:
     max_pin: x
   freetype:
     max_pin: x
-  gstreamer:
-    max_pin: x.x
   icu:
     max_pin: x
   jpeg:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,7 +23,7 @@ source:
     - 0001-shobjidl-Fix-compile-guard-around-SHARDAPPIDINFOLINK.patch  # [win]
 
 build:
-  number: 1012
+  number: 1013
   skip: True  # [ win]
   detect_binary_files_with_prefix: true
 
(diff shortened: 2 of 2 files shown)


croth1:
Wow, conda went back straight to the last ice age to dig out this amazing curl version: 🙄 

`curl                      7.49.1                        0    conda-forge`

looooo:
curl 7.49.1 0 conda-forge has no openssl version specified. So it's the newest conda-forge curl-package that should work with openssl 1.1.1 (but actually does not). Maybe this old curl-package should be labeled "broken"

ccordoba12:
So, is Qt broken due to the update to Gstreamer to 1.14? Or why is this rebuild needed?

croth1:
gstreamer 1.12 didn't get the glib 2.58 upgrade, and we decided to advance the pin here: https://github.com/conda-forge/gstreamer-feedstock/issues/23

ccordoba12:
Well, I disagree with @scopatz because it'd be far more easier (from the Qt point of view) to simply recompile Gstreamer 1.12 with the new Glib since that doesn't require a rebuild here.

@scopatz, what do you say?

ccordoba12:
Besides, this Qt version is very old so it could not be compatible with Gstreamer 1.14

croth1:
@ccordoba12 @scopatz  @isuruf, @conda-forge/core we should really, really start drafting a common strategy on how pins are chosen and advanced. Things are confusing and messy right now.

croth1:
1.14 is very likely to be compatible because macOS already builds with 1.14. 

Rebuilding 1.12 will require at least some work, because gstreamer and gst-plugin-base were merged some time after 1.12 and there is no LTS branch for the 1.12.x series.

scopatz:
There's a lot of confusion about which ABI gstreamer earlier versions was built with. It is probably easier to just rebuild.

ccordoba12:
> 1.14 is very likely to be compatible because macOS already builds with 1.14.

Good point. Then you'll haver to sort the difficulties on Linux, sorry.

ccordoba12:
Ok, this looks good to go. Thanks @isuruf for your help with the last commit.

PR #105, opened 2019-05-23
Fix Qt not being able to create an XKB context

ccordoba12:
Checklist
* [x] Used a fork of the feedstock to propose changes
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Fixes #104 
Fixes https://github.com/conda-forge/pyqt-feedstock/issues/52

@ocefpaf, please help us to recompile and upload a new Linux version.

PR #106, opened 2019-05-24
Re build

ocefpaf:
Alternative to #105 

@ccordoba12 this is the recipe I'm building and I'll upload it as soon as it completes. I'm sending this PR just to make sure we have a direct connection to what was built and the recipe in the feedstock. Note that this has the changes from #105 plus the extra `yum_requirements.txt`.

### DIFF
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -12,7 +12,7 @@ jobs:
     matrix:
       win_cxx_compilervs2015vc14:
         CONFIG: win_cxx_compilervs2015vc14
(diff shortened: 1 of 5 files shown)


ocefpaf:
Tested it locally and it works :tada:

Should close https://github.com/conda-forge/pyqt-feedstock/issues/52 and #105

ocefpaf:
We don't need to wait for the CIs here. I'll cancel the macOS build b/c we don't need a new package there, and Linux and Windows won't build due to time limit.

ccordoba12:
Thanks @ocefpaf!! Great news!

PR #115, opened 2019-08-14
Drop zip source archive

jschueller:
I wonder why we use the zip archive, tar.xz works for win32 too for pyside2, even with patching.
If it succeeds it would allow for the bot to update cleanly the checksum.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -10,10 +10,8 @@ package:
   version: {{ version }}
 
 source:
-  - url: http://download.qt.io/official_releases/qt/5.12/{{ version }}/single/qt-everywhere-src-{{ version }}.tar.xz  # [not win]
-    md5: 6a37466c8c40e87d4a19c3f286ec2542  # [not win]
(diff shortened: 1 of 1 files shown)


mingwandroid:
> I wonder why we use the zip archive, tar.xz works for win32 too for pyside2

The explanation is simple; it is historical. In order to support xz (and various other compression formats) linking against the Windows MSVC2008 CRT I had to do literally months of hard work on the packages themselves and also c99-to-c89.

The completion of this work was adding libarchive to conda-build so that it can decompress basically anythin, then the icing on the cake is the `.conda` format which uses all this to compress and speed the heck out of things.

mingwandroid:
> I wonder why we use the zip archive, tar.xz works for win32 too for pyside2

The explanation is simple; it is historical. In order to support xz (and various other compression formats) linking against the Windows MSVC2008 CRT I had to do literally months of hard work on the packages themselves and also c99-to-c89.

The completion of this work was adding libarchive to conda-build so that it can decompress basically anythin, then the icing on the cake is the `.conda` format which uses all this to compress and speed the heck out of things.

jschueller:
ok, thanks a lot!

PR #120, opened 2019-08-25
Continuation of @jschuelller's 5.12.4 PR.

mingwandroid:
Checklist


ocefpaf:
@jschueller and @mingwandroid please let me know when O should try to build these locally for the upload.

recipe/meta.yaml
```diff
@@ -61,7 +78,8 @@ source:
     folder: opengl32sw                                                                                            # [win64]
 
 build:
-  number: 0
+  number: {{ build_number }}
+  string: {{ build_number }}{{ dev_nature }}
```
isuruf: Use the pkg hash here.

recipe/conda_build_config.yaml
```diff
@@ -1,2 +1,4 @@
+CONDA_BUILD_SYSROOT:       # [osx]
+  - /opt/MacOSX10.12.sdk   # [osx]
```
isuruf: Remove these 2 lines

mingwandroid: We need these for Anaconda Distro, are they harmful?

recipe/meta.yaml
mingwandroid: For my debugging purposes I need to switch this depending on `dev_nature`.

mingwandroid: i.e `{% if dev_nature != '' %}` .. around it or something like that. `dev_nature` will not be set unless forced via the command line.

recipe/conda_build_config.yaml
isuruf: Yes, it's harmful for conda-forge as we are installing the SDK somewhere else.

recipe/meta.yaml
isuruf: I meant use the pkg hash in addition. Something like `{{ PKG_HASH }}_{{ build_number }}{{ dev_nature }}`

recipe/conda_build_config.yaml
mingwandroid: OK, I'm probably OK with aligning our location then. We do need to force 10.12 somehow though. What is the default ATM? Any clues how to force 10.12?

recipe/meta.yaml
mingwandroid: OK, cool

recipe/conda_build_config.yaml
isuruf: It is set to `$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk` which wade sense when we were using system compilers.

Default is 10.9. If you set MACOSX_DEPLOYMENT_TARGET to 10.12 as you've done here in `recipe/conda_build_config.yaml`, the CI script will read that variable, download 10.12, extract it and automatically set `CONDA_BUILD_SYSROOT`. So there's no need to set `CONDA_BUILD_SYSROOT` in the recipe. Just `MACOSX_DEPLOYMENT_TARGET` would do.

jschueller:
@isuruf, your commit fails:
```
cp: ../licenses/LICENSE.FDL: No such file or directory
```

ccordoba12:
Sure, I think so too.

PR #175, opened 2020-12-15
Try qt6

jschueller:
@conda-forge-admin please rerender


jschueller:
@conda-forge-admin please rerender

jschueller:
@conda-forge-admin please rerender

traversaro:
I am not sure if this is the right place to discuss about this or a separate discussion could have a better visibility, anyhow let's write it down here.

Regarding the Qt6 migration, I wonder if it could make sense to use this occasion to shift to have multiple recipes for each qt library? Something like what is done on arch or vcpkg for example, so a [`qt6-base`](https://archlinux.org/packages/extra/x86_64/qt6-base/) recipe, a [`qt6-declarative`](https://archlinux.org/packages/extra/x86_64/qt6-declarative/), and so on so forth. 

This could have some main benefits: 
* It should reduce the risk of the build taking more time that the max time permitted by the CI services, something that as far as I under is happens quite often now, leaving to the maintainers the relatively complicated task of manual local builds and manual uploads
* It would permit to simplify adding support for new platforms that may not support all of qt, for example if a new platform support for it can be added incrementally first just on `qt6-base`, and only later for the other qt parts. 

Furthermore, by properly setting the `INSTALL_***DIR` variables and thanks to the versioned names of Qt6 libraries (see https://github.com/archlinux/svntogit-packages/blob/packages/qt6-base/trunk/PKGBUILD#L42 and https://archlinux.org/packages/extra/x86_64/qt6-base/files/) it is perfectly possible to install side by side qt5 and qt6 packages in the same environment. This would permit to start adding incrementally in staged-recipes the qt6 libraries as they are ready, and only once the overall conda-forge environment is ready switch the qt package to be just a metapackage that depends on the various ` qt*-<libraryname>` packages.  

If there is interest/consensus around this strategy I would be happy to work on starting working on a recipe for qt6-base in staged-recipes. 

Any comment or feedback is welcome @conda-forge/qt @Tobias-Fischer @wolfv

Tobias-Fischer:
Sounds like a great proposal to me @traversaro!

wolfv:
It sounds like a great idea to me. Debugging the Qt recipe is really difficult currently as the build takes so long and ccache or similar tools are not enabled. So splitting this up would be greatly appreciated!

Tobias-Fischer:
Just had a thought.. will we need separate feedstocks as well? Or is there a way to get multiple jobs running, one per output? If all outputs are built in the same job, it is still likely to time out isn't it?

traversaro:
> 
> 
> Just had a thought.. will we need separate feedstocks as well? Or is there a way to get multiple jobs running, one per output? If all outputs are built in the same job, it is still likely to time out isn't it?

I was thinking indeed of having multiple feedstocks, i.e. `qt6-base-feedstock`, `qt6-declarative-feedstock`, etc etc.

h-vetinari:
> 
> 
> Just had a thought.. will we need separate feedstocks as well? Or is there a way to get multiple jobs running, one per output? If all outputs are built in the same job, it is still likely to time out isn't it?

Sounds like you might also be interested in this idea: https://github.com/conda-forge/conda-smithy/issues/1472 🙃

jschueller:
@conda-forge-admin please rerender

jschueller:
I'm moving this to https://github.com/conda-forge/staged-recipes/pull/16741
For now it builds base+svg and succeeds on linux&osx

PR #203, opened 2021-09-13
Remove useless and error-ing call to `pkg-config --pclibdir`

crisbal:
The rationale behind this removal is the following:

* `--pclibdir` is a non-existent flag of `pkg-config`, resulting in a not 0 exit code (I can't find a reference to this `--pclibdir` existing ever, except for this line in this repo)

* Due to how the Bash code is written, the fail is not passed to the calling script (`build.sh`). Infact `export MYVAR=$(error-now)` will always be true.  If the line is rewritten as `MYVAR=$(error-now); export MYVAR` the call will fail. 
Infact changing this line to `PKG_CONFIG_LIBDIR=$("${USED_BUILD_PREFIX}"/bin/pkg-config --pclibdir); export PKG_CONFIG_LIBDIR` will fail exactly with `--pclibdir unknown option`.
For more details on this see the following: https://github.com/koalaman/shellcheck/wiki/SC2155

Effectively this line does nothing and an error is exposed when it is rewritten following best-practices.

**Checklist**
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -5,7 +5,7 @@
 jobs:
 - job: linux
   pool:
-    vmImage: ubuntu-16.04
+    vmImage: ubuntu-latest
   strategy:
     matrix:
       linux_64_:
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -86,7 +86,7 @@ jobs:
 
     - script: |
         call activate base
-        conda.exe build "recipe" -m .ci_support\%CONFIG%.yaml
+        conda.exe build "recipe" -m .ci_support\%CONFIG%.yaml --suppress-variables
       displayName: Build recipe
       env:
         PYTHONUNBUFFERED: 1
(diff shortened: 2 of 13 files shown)


crisbal:
@conda-forge-admin, please rerender

crisbal:
To add a bit more context to this, in my current company we have a fork of this package and while running `sheellcheck` on the `build.sh` script the error about the dangerous use of `export VAR=$(error-now)` was reported. Upon fixing it to `VAR=$(error-now);export VAR` the build started failing because of `--pclibdir`. 

Having it the way it is now it does not change any behavior, but to me it looks like code that has no purpose. It is definitely possible I am missing something there, so don't take my word for granted.

ccordoba12:
Yep, let's close it.

@crisbal, if you still need this, please submit it to the qt-main feedstock. And sorry for not taking your work into account before but the team was focused on updating Qt to 5.15.

PR #227, opened 2022-04-21
Qt main and Qt 5.15.2

hmaarrfk:
I think that for now, we should keep this as an arch dependent package so we can test that compilation works for all packages

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)


hmaarrfk:
@conda-forge-admin please rerender

hmaarrfk:
Not sure why it isn't picking up the patched version. We can try again tomorrow.

hmaarrfk:
@conda-forge/qt-main @conda-forge/qt this is ready for review.

hmaarrfk:
5.15.3 will have to wait for: https://github.com/conda-forge/qt-webengine-feedstock/pull/5

izahn:
@conda-forge/qt can we get a review here? We've been waiting a long time for `qt=5.15`, and several packages need it before they can be updated.

recipe/meta.yaml
```diff
@@ -1,196 +1,24 @@
-# Note, you cannot build this with macOS SDK < 10.12, despite there being a few patches
-# to add back support. QtWebengine scuppers us with:
-#   error: unknown type name 'SecKeyAlgorithm'
-# .. these 10.11 support patches should be removed anyway as they could break some things.
-# .. such as pop-up Window mouse-events.
-{% set version = "5.12.9" %}
+# This has become a metapacakge for qt-base and other related packages
+{% set version = "5.15.2" %}
```
isuruf, line 2: Why 5.15.2 when there's 5.15.3?

isuruf:
> https://github.com/orgs/conda-forge/teams/qt can we get a review here? We've been waiting a long time for qt=5.15, and several packages need it before they can be updated.

We'll probably not use `qt` as the package just like we don't use `matplotlib` and just use `matplotlib-base`.

Please send a PR to the bot to add a migrator to change `qt` to `qt-main` in recipes.

recipe/meta.yaml
hmaarrfk: we need to finish the icu70 migration correctly before we can do 5.15.3

izahn:
> We'll probably not use `qt` as the package just like we don't use `matplotlib` and just use `matplotlib-base`.
> 
> Please send a PR to the bot to add a migrator to change `qt` to `qt-main` in recipes.

I've never done this before, and it looks like name migrations are rare. Is there an example I can look at? Or maybe someone with more experience can handle this one.

Tobias-Fischer:
Here is an example: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/matplotlib_base.py and some documentation is here: https://regro.github.io/cf-scripts/migrators.html#custom-migrations

But it looks like this one is more complicated than it should be because something went wrong during the migration and half of the file is about the fix of that rather than the actual migration ..

izahn:
> Here is an example: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/matplotlib_base.py and some documentation is here: https://regro.github.io/cf-scripts/migrators.html#custom-migrations
> 
> But it looks like this one is more complicated than it should be because something went wrong during the migration and half of the file is about the fix of that rather than the actual migration ..

Oh my, yeah, this looks like it needs attention from someone with more experience with migrations then I have.

Tobias-Fischer:
Hi @beckermr - sorry to bother you, but do you have an idea how we can handle the migration from `qt 5.12` to `qt-main 5.15`? It seems like it's going beyond the expertise of most of us, there doesn't seem to be an example that can be easily adapted (the `matplotlib` one I linked above does a rename but no version bump, and contains some code fixing a broken migration) and you seem to be very active in https://github.com/regro/cf-scripts/tree/master/conda_forge_tick/migrators so I thought I'd reach out.

isuruf:
You need to write a mini-migrator that changes qt to qt-main and add that as a piggy-back migrator in the version migration at https://github.com/regro/cf-scripts/blob/64c14692b14cf07655b65d01ec8fd75649e56ac5/conda_forge_tick/auto_tick.py#L661.
Then, you start a regular migration yaml like https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/migrations/icu70.yaml. A sample mini-migrator is at https://github.com/regro/cf-scripts/blob/64c14692b14cf07655b65d01ec8fd75649e56ac5/conda_forge_tick/migrators/cross_compile.py#L37

hmaarrfk:
Here is my attempt:

https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2929

Tobias-Fischer:
Should we update this PR to use 5.15.3 instead, now that https://github.com/conda-forge/qt-webengine-feedstock/pull/5 is merged?

hmaarrfk:
Having both versions wouldn't be a bad idea. we can deal with that later. Some stuff is still on ICU69. Seeing as this package is now trivial, it isn't a big deal.

PR #230, opened 2022-06-05
Update version

hmaarrfk:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 # This has become a metapacakge for qt-base and other related packages
-{% set version = "5.15.2" %}
+{% set version = "5.15.3" %}
 
 package:
   name: qt


hmaarrfk:
I think some old environment are failing to "update" because `qt` is at version `5.15.2` but their packages requires some migrations that are only available with 5.15.3 (which is why I was really adamant about holding off on 5.15.3 during the initial rollout).

I'm not sure if there is a way we can tell conda to `forget` about the `qt`, and replace with `qt-main`. But that seems hard.

Instead, I think we should make it an error to depend on `qt` in any new recipe, that way people get the correct pinnings.

jakirkham:
Could add a lint to conda-smithy

hmaarrfk:
???

jakirkham:
Sorry for the lack of clarity, was responding to this last line

> Instead, I think we should make it an error to depend on qt in any new recipe, that way people get the correct pinnings.

We have a number of lints [starting here]( https://github.com/conda-forge/conda-smithy/blob/566467daabe5ce1659ffef762fddbdaf0873bf2d/conda_smithy/lint_recipe.py#L144 ). For example [this `pip` hint]( https://github.com/conda-forge/conda-smithy/blob/566467daabe5ce1659ffef762fddbdaf0873bf2d/conda_smithy/lint_recipe.py#L615 ) could be a good starting point for a `qt` related lint (or hint)

